PR #3, opened 2019-09-19
Fix touch input rotation / Disable single-touch mode

StollD:
Previous discussion:
* https://github.com/jakeday/linux-surface/issues/549

Currently, IPTS exposes two seperate touchscreen devices: One for single-touch and one for multi-touch. However, this confuses window managers such as GNOME's mutter, who cannot know which one is active at the moment. The result of that is that mutter will always rotate the single-touch device. The multi-touch one isn't rotated properly, even though multi-touch mode is whats probably being used in most cases. I am not sure if anyone actually uses single-touch mode to be honest.

Since the original IPTS implementation modified hid-multitouch to make single-touch mode work [1], reverting those changes disables single-touch, and fixes the touch input rotation in mutter. (Courtesy of @kitakar5525). This has the additional benefit of reducing the amout of changes to external code that IPTS makes. The alternative is to expose it as a stylus, which doesn't really work that well, and still requires patches to hid-multitouch.

Since the APIs to switch between multi-touch and single-touch mode are pretty much useless now, the second commit removes them. Internally though, IPTS still has code to handle mode switching since that is part of the initialisation process as it seems.

* [1] https://github.com/ipts-linux-org/ipts-linux-new/commit/7cc1f5419bb16ad6584d3ddac9e8b62ea2a8b5f7

### DIFF
--- a/drivers/hid/hid-multitouch.c
+++ b/drivers/hid/hid-multitouch.c
@@ -169,7 +169,6 @@ struct mt_device {
 static void mt_post_parse_default_settings(struct mt_device *td,
 					   struct mt_application *app);
 static void mt_post_parse(struct mt_device *td, struct mt_application *app);
-static int cc_seen = 0;
 
 /* classes of device behavior */
 #define MT_CLS_DEFAULT				0x0001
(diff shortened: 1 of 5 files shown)


qzed:
Thanks!

kitakar5525:
My little concern is that we cannot distinguish touch devices by name anymore:
```bash
# before
$ libinput list-devices | grep ipts
Device:           ipts 1B96:005E UNKNOWN
Device:           ipts 1B96:005E
Device:           ipts 1B96:005E Touchscreen
Device:           ipts 1B96:005E Mouse

# after
$ libinput list-devices | grep ipts
Device:           ipts 1B96:005E
Device:           ipts 1B96:005E
```

No impact on functionality, though.
I hope we can change the name by IPTS driver side rather than patching to hid-multitouch.c at some point, like `ipts 1B96:005E Touchscreen` and `ipts 1B96:005E Pen`.

(The `ipts 1B96:005E UNKNOWN` device was a single-touch device.)
(I don't know what was the purpose of the `ipts 1B96:005E Mouse`. It seems to be not used by any input for me.)

qzed:
@kitakar5525 Thanks, that's a valid concern. I've moved this to a new issue (#4).

PR #5, opened 2019-09-22
Add secureboot pre-signing to the kernel

StollD:
This adds the ability for you and other packagers to pre-sign the kernels for secureboot. 

The kernel build system will look for two files:
* `keys/MOK.key` (private key)
* `keys/MOK.crt` (public key in PEM format)

and sign the produced vmlinux / bzImage file with it. If no keys are found, it will be silently ignored and the kernel will be unsigned.

Adding it to the build system avoids having to wrap everything. Right now you need to first to `make all`, then sign the image, and then do something like `make bindeb-pkg`. With this, you can just add your keys to the kernel source, and run `make deb-pkg`.

Users then only need to be able to download the public key (in DER format), and enroll it into `shim` to be able to boot the kernel - instead of generating the keys and signing the kernel all on their own. 

We could also edit the Arch / Debian packaging templates to install the key automatically (like [this](https://github.com/StollD/fedora-linux-surface/blob/master/pkg/linux-surface-secureboot/linux-surface-secureboot.spec#L30-L33)), or package them seperately if we ever get to setting up a package repository for this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,6 +132,9 @@ signing_key.priv
 signing_key.x509
 x509.genkey
 
+# Secureboot certificate
+/keys/
+
 # Kconfig presets
 /all.config
 /alldef.config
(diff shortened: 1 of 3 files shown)


qzed:
Thanks!

PR #7, opened 2019-10-03
Add firmware declarations for ipts_surface

StollD:
Declaring the required firmware files as MODULE_FIRMWARE makes it possible to include the driver in the initramfs, in case that is needed or desired.

Even though there are files that are loaded through a config file, that config file is the same for every Surface using IPTS, so this should be fine. (This might also change later in case we decide to remove the config file and move it's functions into the companion driver)


qzed:
Thanks!

PR #11, opened 2019-10-25
Move ipts_fw_config.bin into the companion driver

StollD:
As discussed on IRC, this modifies IPTS to have the companion driver provide the firmware config directly, instead of loading it from an obscure binary file that was provided by Intel ages ago.

To archive this the companion driver interface has been overhauled to be more easily extensible (just add the things you need to the `ipts_companion` struct and they will be usable from IPTS).

The public header files for IPTS under `include/linux` have been updated as well, since it was neccessary to add another one there (`ipts-binary.h`, formerly `ipts-binary-spec.h` in the driver).


qzed:
CC @GrayHatter, if you want to look at it

drivers/misc/ipts/companion/ipts-surface.c
```diff
@@ -1,44 +1,69 @@
...
+	.num_of_data_files = 3,
+	.data_file = {
+		{
+			.io_buffer_type = IPTS_CONFIGURATION,
+			.flags = IPTS_DATA_FILE_FLAG_NONE,
+			.file_name = "config.bin",
+		},
+
+		// The following files are part of the config, but they don't
+		// exist, and the driver never requests them.
```
GrayHatter: If this is true, can we comment these both out and prove that?

drivers/misc/ipts/companion/ipts-surface.c
```diff
@@ -49,11 +74,11 @@ static int ipts_surface_probe(struct platform_device *pdev)
...
 	}
 
-	ret = intel_ipts_add_fw_handler(&ipts_surface_request_firmware,
-		(void *)acpi_device_hid(adev));
+	ipts_surface_companion.data = (void *)acpi_device_hid(adev);
+	ret = ipts_add_companion(&ipts_surface_companion);
 	if (ret) {
-		dev_info(&pdev->dev, "Adding IPTS firmware handler failed, "
-			"error: %d\n", ret);
+		dev_info(&pdev->dev, "Adding IPTS companion failed, "
```
GrayHatter: why is this info and not error?

drivers/misc/ipts/ipts-companion.c
```diff
@@ -0,0 +1,203 @@
...
+#define IPTS_FW_PATH_FMT "intel/ipts/%s"
+#define IPTS_FW_CONFIG_FILE "ipts_fw_config.bin"
+
+ipts_companion_t *ipts_companion;
+DEFINE_MUTEX(ipts_companion_lock);
+
+bool ipts_companion_available(void)
+{
+	bool ret;
+	mutex_lock(&ipts_companion_lock);
```
GrayHatter: is there a reason you picked mutex over spin here?

drivers/misc/ipts/ipts-companion.c
```diff
@@ -0,0 +1,203 @@
...
+ */
+
+int ipts_add_companion(ipts_companion_t *companion)
+{
+	int ret = 0;
+	mutex_lock(&ipts_companion_lock);
+
+	if (ipts_companion != NULL) {
+		ret = -EBUSY;
+		goto add_companion_return;
```
GrayHatter: This looks awkard, can you rewrite it without the goto?

drivers/misc/ipts/ipts-companion.c
```diff
@@ -0,0 +1,203 @@
...
+	return ret;
+}
+EXPORT_SYMBOL_GPL(ipts_add_companion);
+
+int ipts_remove_companion(ipts_companion_t *companion)
+{
+	int ret = 0;
+	mutex_lock(&ipts_companion_lock);
+
+	if (ipts_companion == NULL || companion == NULL) {
```
GrayHatter: check `companion != NULL` before getting a mutex

drivers/misc/ipts/ipts-companion.c
```diff
@@ -0,0 +1,203 @@
...
+	ipts_companion = companion;
+
+add_companion_return:
+
+	mutex_unlock(&ipts_companion_lock);
+	return ret;
+}
+EXPORT_SYMBOL_GPL(ipts_add_companion);
+
+int ipts_remove_companion(ipts_companion_t *companion)
```
GrayHatter: shouldn't this function be in the header as well?

drivers/misc/ipts/ipts-companion.c
```diff
@@ -0,0 +1,203 @@
...
+		for (j = 0; j < fw[i]->num_of_data_files; j++) {
+			*(ipts_bin_data_file_info_t *)itr = fw[i]->data_file[j];
+			itr += sizeof(ipts_bin_data_file_info_t);
+		}
+	}
+
+	return fw_list;
+}
+
+int ipts_request_firmware_config(ipts_info_t *ipts, ipts_bin_fw_list_t **cfg)
```
GrayHatter: this whole function can be written better without any gotos

drivers/misc/ipts/ipts-companion.c
```diff
@@ -0,0 +1,203 @@
...
+	snprintf(fw_path, MAX_IOCL_FILE_PATH_LEN, IPTS_FW_PATH_FMT, name);
+	ret = request_firmware(fw, fw_path, device);
+
+request_firmware_return:
+
+	mutex_unlock(&ipts_companion_lock);
+	return ret;
+}
+
+ipts_bin_fw_list_t *ipts_alloc_fw_list(ipts_bin_fw_info_t **fw)
```
GrayHatter: can this be static?

include/linux/ipts-companion.h
```diff
@@ -0,0 +1,22 @@
+#ifndef _INTEL_IPTS_COMPANION_H_
```
GrayHatter: this ifdef doesn't match the file name, and it's name does have a duplicate can these two be merged down to one?

include/linux/ipts-binary.h
```diff
@@ -81,38 +84,62 @@ typedef enum {
...
-	unsigned int mem_offset;
+	u32 imm_offset;
+	u32 mem_offset;
 } ipts_bin_bufid_patch_t;
 
+typedef enum {
+	IPTS_DATA_FILE_FLAG_NONE,
+	IPTS_DATA_FILE_FLAG_SHARE,
+	IPTS_DATA_FILE_FLAG_ALLOC_CONTIGUOUS,
+} ipts_bin_data_file_flags_t;
```
GrayHatter: If you want these to be bit flags, they should have explicit values.

Unless this is something hidden in the firmware, and if that's true, we should document that because... ow!

drivers/misc/ipts/companion/ipts-surface.c
StollD: I would rather keep them, as a reference, so that we don't have to reverse engineer the binary file again at some point. That they aren't requested can already be proven, because the files don't exist, and there are no errors about them missing in the log.

drivers/misc/ipts/companion/ipts-surface.c
StollD: Because it could happen that multiple companions get probed, or the user explicitly tells IPTS to ignore the companion. In that case, we didn't really want the companion to error.

drivers/misc/ipts/ipts-companion.c
StollD: Because qzed told me so when writing the original implementation :P (in `ipts-fw.c` which was deleted).

I am not that familiar with synchronisation inside of the kernel, so I just kept it like that.

drivers/misc/ipts/ipts-companion.c
StollD: I could but then I would have to duplicate the `mutex_unlock`. I'd rather keep it that way, the kernel coding guidelines explicitly recommend goto in such cases.

drivers/misc/ipts/ipts-companion.c
StollD: Both, `ipts_add_companion` and `ipts_remove_companion` are exported in the public header file (`include/linux/ipts-companion.h`) for other modules to use. The other functions are internal use by IPTS itself, so they are exported in the local header file inside `driver/`

drivers/misc/ipts/ipts-companion.c
StollD: Well, the mutex was added to make sure that every access to the companion driver is synchronised, so I am not sure if that wouldn't defeat the purpose of the mutex. 

@qzed ?

drivers/misc/ipts/ipts-companion.c
StollD: Probably. I didn't add it because when I wrote it I still wasnt sure it I had to export it to the companion driver or not, but it ended up being an internal function.

drivers/misc/ipts/ipts-companion.c
StollD: See my answer above: Not using gotos would yield duplicated `mutex_unlock` calls all over the place, and using goto in such a case is what the kernel coding guidelines advise.

include/linux/ipts-binary.h
StollD: Their values should be 0, 1, and 2, which would already make them a bitflag. I agree it would be neccessary if we ever add another element to it (which will never happen probably because those come from the firmware)

include/linux/ipts-companion.h
StollD: Yes the header guards are a total chaos in IPTS (everywhere, not just in this case). I would rather keep it that way now and fix them in a broader cleanup at some point though.

Regarding the duplicated filename: There is a second `ipts-companion.h` but I would like to keep them seperate because they do different things: The one in `include/` is everything that goes from the companion to IPTS. The other one is the opposite, it contains everything that goes from IPTS to the companion.

drivers/misc/ipts/companion/ipts-surface.c
qzed: I think we should keep them for now. We can always remove them in a future commit, but it's probably a good idea to have them in the history.

drivers/misc/ipts/companion/ipts-surface.c
qzed: I think multiple companions probing should be at least a warning, if not an error. I'd normally expect that only one companion driver gets loaded for a specific machine. Also, as far as I can tell, if a user sets the modparam to ignore the IPTS companion, `ipts_add_companion` does not return an error, so I think it's safe to change this to something like `dev_warn(..., "IPTS companion driver already present, aborting\n")`.

drivers/misc/ipts/ipts-companion.c
qzed: My reasoning was mostly: Mutex is the standard tool and it's initialization code, so we don't need the performance a spin-lock would give us here. Also the lock is being held during IO operations (`request_firmware`), so not really small critical sections.

drivers/misc/ipts/ipts-companion.c
qzed: I guess in this case it could be rewritten without gotos (as it's pretty small), but for longer functions that's the usual kernel style.

drivers/misc/ipts/ipts-companion.c
qzed: Given that we are on x86, accessing `ipts_companion` outside of the critical section _should_ be safe in this context, but I don't like relying on platform-specifics, especially if unmarked. Also can we guarantee that this will only run on x86? It's IPTS so probably.... Anyways, given that this function is not being called frequently (only during initialization), I don't think optimizations like these make much sense.

drivers/misc/ipts/ipts-companion.c
qzed: Agreed, this should be static.

drivers/misc/ipts/ipts-companion.c
qzed: I'm not necessarily for throwing out the gotos, but I think we can reduce the code beneath the mutex. I think actually only the code until (including) `*cfg = ipts_alloc_fw_list(ipts_companion->firmware_config);` needs to be synchronized. `ipts_request_firmware` does its own checks and synchronization, and the rest is stuff on data that we own at that point.

qzed: Also with that you'll probably get rid of one goto.

include/linux/ipts-companion.h
```diff
@@ -0,0 +1,22 @@
...
+#define _INTEL_IPTS_COMPANION_H_
+
+#include <linux/firmware.h>
+#include <linux/ipts-binary.h>
+
+typedef struct ipts_companion ipts_companion_t;
+
+typedef int (*ipts_fw_handler_t)(const struct firmware **, const char *,
+		struct device *, ipts_companion_t *companion);
```
qzed: Since we're going over this again, I'd kind of prefer having the `companion` as the first parameter. Makes the association a bit clearer (as in "for companion x load firmware with parameters y, z, ...").

include/linux/ipts-companion.h
qzed: I agree on the separation part, but the names being equal could cause some confusion. I don't have any suggestion for better names though.

drivers/misc/ipts/companion/ipts-surface.c
StollD: Done, changed `dev_info` to `dev_warn`.

drivers/misc/ipts/ipts-companion.c
StollD: Done, used if / else instead of goto. The if condition looks a bit janky, but personally I'm fine with that.

drivers/misc/ipts/ipts-companion.c
StollD: I looked into how to reduce the amount of gotos in that function, but I found no nice way that allows for all the fallback behaviour that is there, and that doesn't require duplicating the call to `mutex_unlock`.

If you *want* me to change it, I will look into it again ofc.

drivers/misc/ipts/ipts-companion.c
StollD: Done.

include/linux/ipts-companion.h
StollD: Done.

drivers/misc/ipts/ipts-companion.c
qzed: What I meant was something like this:

```c
int ipts_request_firmware_config(ipts_info_t *ipts, ipts_bin_fw_list_t **cfg)
{
	int ret = 0;
	const struct firmware *config_fw = NULL;
	mutex_lock(&ipts_companion_lock);

	// Check if a companion was registered. If not, skip
	// forward and try to load the firmware config from a file
	if (ipts_modparams.ignore_companion || ipts_companion == NULL) {
		mutex_unlock(&ipts_companion_lock);
		goto config_fallback;
	}

	if (ipts_companion->firmware_config != NULL) {
		*cfg = ipts_alloc_fw_list(ipts_companion->firmware_config);
		mutex_unlock(&ipts_companion_lock);
		return 0;
	}

config_fallback:

	// If fallback loading for the firmware config was disabled, abort.
	// Return -ENOENT as no config file was found.
	if (ipts_modparams.ignore_config_fallback) {
		ret = -ENOENT;
		goto config_return;
	}

	// No firmware config was found by the companion driver,
	// try loading it from a file now
	ret = ipts_request_firmware(&config_fw, IPTS_FW_CONFIG_FILE,
			&ipts->cldev->dev);
	if (ret)
		return ret;

	*cfg = (ipts_bin_fw_list_t *)config_fw->data;
	release_firmware(config_fw);
	return 0;
}
```

You only need the mutex to guard the direct `ipts_companion` access, the rest doesn't need to be in the critical section (especially as `ipts_request_firmware` does all that itself). I'd still keep the goto for the fallback as I don't see a cleaner option.

StollD: Ok, yeah, that is definitly nicer. Updated the function, thank you.

include/linux/ipts-companion.h
StollD: What we could do is remove the `ipts` prefix from all the files under `driver/misc/ipts`. You know they are for IPTS through the directory name anyways. Their names would be similar to the ones under `include/` but they wouldn't be the same.

That is totally out of scope for this pull request though imo.

qzed: That might be a solution. Definitely agreed on the out-of-scope part.

include/linux/ipts-binary.h
qzed: Yeah I think those should nevertheless be explicit, especially to document that this is not some random enum that could be reordered but the values are fixed by the firmware.

StollD: Updated, but force pushing confused github so it still shows the old code for me here. The full diff on the PR works though.

PR #14, opened 2019-11-07
Replace the HID patch with a better one that fixes all PTP touchpads

archseer:


qzed:
As per https://github.com/qzed/linux-surfacegen5-acpi/pull/24.

On it's way upstream.
Patchwork: https://patchwork.kernel.org/patch/11230017/

PR #15, opened 2019-11-07
IPTS code cleanup

StollD:
As I mentioned on IRC before, I have been working on cleaning up the code for IPTS. The main goal was to make every file pass the `checkpatch` utility that comes with the kernel and checks for styleguide violations, and for the code to look better and more consistent.

Main changes:
* Every file that is part of the IPTS patch passes `checkpatch` (with two exceptions but those are edge cases not covered by `checkpatch`)
* Fixed all the header guards to be consistent
* Removed most / all `intel_ipts_` prefixes, in favor of using `ipts_` consistently.
* Removed the `ipts-` prefix from the files under `driver/misc/ipts`
* Moved the IPTS debugging options into a module parameter, so they can be enabled dynamically (`intel_ipts.debug` and `intel_ipts.debug_thread`)
* Documented the output from the IPTS debug thread

For the most part these are just semantic patches, I didn't change the logic anywhere - except for some smaller cases where I inverted an if or broke up multiple ifs into gotos to stay within the 80 char / line limit that checkpatch expects.

I have been running these changes on my surface for the past week or so now, and there are no regressions introduced by this.


drivers/misc/ipts/companion.c
```diff
@@ -1,27 +1,37 @@
+// SPDX-License-Identifier: GPL-2.0-or-later
+/*
+ *
+ * Intel Precise Touch & Stylus
+ * Copyright (c) 2016 Intel Corporation
```
StollD: Note: I added the intel copyright header to all of the files, even the ones that were added later on.

Having some files with the header, and some files without looked dumb to me somehow, and Intel provided the biggest part of the driver and the technology. As far as "my" files are concerned (I *think* I am the only one who ever added files tbh.), they are mostly just other code from Intel moved around anyways.

Of course we could and should add other names there if we decide to. Or maybe just add an entry for `linux-surface` as whole.

But I just wanted to mention it in case someone got confused about why I added those.

qzed: I'm pretty sure you could pin your name at least on the companion driver. I mean you pretty much wrote every part of it. But we can always add that later.

qzed:
Woah, that looks like it was a whole lot of work! Thanks!

PR #16, opened 2019-11-08
Move no_feedback DMI check to the companion driver

StollD:
Having a DMI match for some surface devices doesn't really match with the concept of having a device-specific companion driver, so this moves the matching code to the companion.

ipts-surface is now setup in a way that allows us to add device-specificn data pretty easily. All you need is the interface to IPTS, which can just be copied from no_feedback.

### DIFF
--- a/drivers/misc/ipts/companion.c
+++ b/drivers/misc/ipts/companion.c
@@ -210,19 +210,19 @@ config_fallback:
 
 }
 
-bool ipts_needs_no_feedback(void)
+unsigned int ipts_get_quirks(void)
 {
-	bool ret;
+	unsigned int ret;
 
 	// Make sure that access to the companion is synchronized
 	mutex_lock(&ipts_companion_lock);
 
 	// If the companion is ignored, or doesn't exist, assume that
-	// the device doesn't need no_feedback enabled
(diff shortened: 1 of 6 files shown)


drivers/misc/ipts/companion.c
```diff
@@ -209,3 +209,22 @@ int ipts_request_firmware_config(struct ipts_info *ipts,
...
+
+	// Make sure that access to the companion is synchronized
+	mutex_lock(&ipts_companion_lock);
+
+	// If the companion is ignored, or doesn't exist, assume that
+	// the device doesn't need no_feedback enabled
+	if (ipts_modparams.ignore_companion || ipts_companion == NULL)
+		ret = false;
+	else
+		ret = ipts_companion->needs_no_feedback(ipts_companion);
```
qzed: I like the idea of moving the device quirks to the companion driver, but can we do that a bit more generic? Currently adding a new quirk would require adding a function to the companion driver interface. I'd prefer having a generic quirk function that returns a `uint64_t` bitfield and something like `#define IPTS_QUIRK_NO_FEEDBACK BIT(0)`.

qzed:
Thanks!

PR #23, opened 2019-12-20
ipts: Dont automatically switch sensor modes

StollD:
We don't want IPTS to set itself to singletouch mode in case it thinks that might fix anything, because it doesn't. Even a malfunctioning multitouch IPTS has a higher chance of working than singletouch.

This could potentially fix some of the stability issues that some models have. Even if not, it is one bit of confusing code less.

Note: I haven't had the time to test this yet, I just ran across it while working on something else. But it shouldn't cause any issues.


qzed:
Thanks, makes sense! I'll test it and merge then.

PR #24, opened 2019-12-26
surface3-spi: workaround: disable DMA mode to avoid crash by default

kitakar5525:
Hello, on Surface 3, we're facing touch input crash with some kernel config setup, especially with your build and Arch Linux stock kernel.

It turned out that DMA mode seems to be broken with surface3-spi touchscreen driver. So, this commit will disable DMA mode by default and use PIO mode instead for the driver.
Any other changes beyond Surface 3 touchscreen are not intended.

If you find something is weird, let me know. Also, this is just a workaround, I should actually fix   this driver with DMA mode if I can. So, If you know how to set up DMA mode, also please let me know.

Link to issue: jakeday/linux-surface#596

Tested on Arch Linux 5.4.1 with Surface 3, which will use DMA by default.
This commit made the driver use PIO by default and no touch input crash.
Also tested on chromeos-4.19 4.19.90 with Surface 3, which will use PIO by default
even without this commit. After this commit, it still uses PIO and confirmed
no functional changes regarding touch input.

### DIFF
--- a/drivers/input/touchscreen/surface3_spi.c
+++ b/drivers/input/touchscreen/surface3_spi.c
@@ -25,6 +25,12 @@
 #define SURFACE3_REPORT_TOUCH	0xd2
 #define SURFACE3_REPORT_PEN	0x16
 
+bool use_dma = false;
+module_param(use_dma, bool, 0644);
+MODULE_PARM_DESC(use_dma,
+				"Disable DMA mode if you encounter touch input crash. "
+				"(default: false, disabled to avoid crash)");
+
 struct surface3_ts_data {
 	struct spi_device *spi;
 	struct gpio_desc *gpiod_rst[2];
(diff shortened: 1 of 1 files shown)


kitakar5525:
Ah, I've forgotten to make a PR for 4.19 branch. Anyway, I don't know how to make a PR to multiple branches (5.3 and 4.19) at the same PR.
Should I make a new PR for 4.19 branch?

qzed:
Thanks! Unfortunately I don't know much about DMA, so I don't have any clue of what could be going wrong.



> Ah, I've forgotten to make a PR for 4.19 branch. Anyway, I don't know how to make a PR to multiple branches (5.3 and 4.19) at the same PR.
> Should I make a new PR for 4.19 branch, too?

No need, I'll update the other branches.

PR #29, opened 2020-01-11
Surface3-OEMB: add quirk for Surface 3 with broken DMI table

kitakar5525:
As a request from a person who is also suffering from "OEMB" issue on IRC channel.
Please port this into also 4.19 and 5.4 kernels when approved this PR, should apply with no conflicts.

The first patch (ASoC) is from Android-x86 project. The second patch is made by me referring to the first patch (they don't have a similar patch for surface3-wmi, so I made it).
I credited the first patch author in commit message.

Fixes broken Sound and surface3-wmi driver probe on the affected systems.

---
On some Surface 3, DMI table gets corrupted for unknown reasons
and breaks existing DMI matching used for device-specific quirks.

This PR adds the (broken) DMI info for the affected systems.

On the affected systems, dmidecode will look like this:
```
        $ sudo dmidecode
        [...]
        BIOS Information
            Vendor: American Megatrends Inc.
        [...]
        System Information
            Manufacturer: OEMB
            Product Name: OEMB
        [...]
```

Expected:
```
        $ sudo dmidecode
        [...]
        BIOS Information
            Vendor: (???, I think something like "Microsoft Corporation")
        [...]
        System Information
            Manufacturer: Microsoft Corporation
            Product Name: Surface 3
        [...]
```

### DIFF
--- a/drivers/platform/x86/surface3-wmi.c
+++ b/drivers/platform/x86/surface3-wmi.c
@@ -37,6 +37,13 @@ static const struct dmi_system_id surface3_dmi_table[] = {
 			DMI_MATCH(DMI_PRODUCT_NAME, "Surface 3"),
 		},
 	},
+	{
+		.matches = {
+			DMI_MATCH(DMI_BIOS_VENDOR, "American Megatrends Inc."),
+			DMI_MATCH(DMI_SYS_VENDOR, "OEMB"),
+			DMI_MATCH(DMI_PRODUCT_NAME, "OEMB"),
+		},
+	},
 #endif
 	{ }
 };
--- a/sound/soc/codecs/rt5645.c
+++ b/sound/soc/codecs/rt5645.c
@@ -3673,6 +3673,15 @@ static const struct dmi_system_id dmi_platform_data[] = {
 		.driver_data = (void *)&intel_braswell_platform_data,
 	},
 	{
+		.ident = "Microsoft Surface 3",
+		.matches = {
+			DMI_MATCH(DMI_BIOS_VENDOR, "American Megatrends Inc."),
+			DMI_MATCH(DMI_SYS_VENDOR, "OEMB"),
+			DMI_MATCH(DMI_PRODUCT_NAME, "OEMB"),
+		},
+		.driver_data = (void *)&intel_braswell_platform_data,
+	},
+	{
 		/*
 		 * Match for the GPDwin which unfortunately uses somewhat
 		 * generic dmi strings, which is why we test for 4 strings.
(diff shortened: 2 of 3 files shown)


kitakar5525:
I used my github name as author name in the first commit (ASoC) because I didn't want to bother the original author with github notifications.
(I've already credited the author name in its commit message.)

And now I learned a bit about how everyone treats the author name on git.
In general, the author name should remain original author name (even when I modified some lines).

```bash
git commit --amend --author='Author Name <email@address.com>'
```

So, I'll change back the author name of the first commit to the original author name by force-pushing

*(I feel I need to learn git)*

kitakar5525:
Changing back the author name done.

kitakar5525:
(And improved commit messages)

qzed:
Thanks! Any idea if this is being pushed upstream?

kitakar5525:
> Any idea if this is being pushed upstream?

I should definitly try now (and also surface3-spi DMA issue: https://github.com/linux-surface/kernel/commit/6c196dee2d1f7be5e5e94a802fdcac23505c4178), although repairing the broken DMI is the best idea if possible.

I don't know where is the right place to discuss. Bugzilla?

qzed:
You should probably ask the original author if he intends of pushing this upstream. I'm not sure if the DMI can be repaired. Also if it can be repaired, it might not be the best option for users. As a note: If you want to get this upstream you may need to change the match a bit as I think `American Megatrends Inc.` and `OEMB` may be too generic (but they'll tell you then what they want to have changed specifically).

Apart from that, I'm not quite sure where this should go, to be honest. Both, bugzilla and mailing list should work. You'll probably get quicker results on the list, but which category. Platform-drivers-x86 probably isn't the right place for DMI in general, but as it affects driver quirks maybe there? Linux-efi maybe as DMI is provided by firmware?

Running `./scripts/get_maintainer.pl drivers/firmware/dmi_scan.c` only gives
```
Jean Delvare <jdelvare@suse.com> (maintainer:DMI/SMBIOS SUPPORT)
linux-kernel@vger.kernel.org (open list)
```
but maybe that's a starting point figuring out where to go from there.

kitakar5525:
Thank you. I'll first contact the author.

andy-shev:
These patches are hacks. They are not going to be upstreamed. Sorry for this news, but we need to understand the root cause before going forward.

kitakar5525:
> These patches are hacks. They are not going to be upstreamed.

I see. Yes, it's a hack. It can't be helped that they are not going to be upstreamed.
I hope there were a way to override DMI table, like overriding osi string (`acpi_osi` parameter). I understand there is no demand to do such a thing, though.

> we need to understand the root cause before going forward.

I can't say for sure, but it might be caused by force power-off by long-pressing power button.

qzed:
@andy-shev Any idea what could cause this? Also if it's broken in/by firmware, what can we do in kernel to fix/avoid this?

abondis:
I am affected by this issue. I tried many ways with my limited knowledge of DMI and tools related, but could not fix the values... I'm not sure how or why it changed and since I can't go back to the original status, I can't even try to reproduce.

Something like that seems like it would have been great https://patchwork.kernel.org/patch/9591817/ but from the discussion it seems this kind of hack is not acceptable for the kernel either ...

PR #34, opened 2020-01-22
v4.19: ipts: Cleanup intel_guc_submission to pass checkpatch

kitakar5525:
After backporting 5.3/5.4 changes, checkpatch added one error compared
to IPTS patch for 5.3:

        ERROR: else should follow close brace '}'
        #386: FILE: drivers/gpu/drm/i915/intel_guc_submission.c:117:
        +	}
        +	else {

This commit fixes the error.

The cause of the new error is that when this error was fixed in a
later kernel, the change was not backported to 4.19. On 5.3, the
error was already fixed when IPTS support (871f089) was added.


qzed:
Thanks!

PR #35, opened 2020-01-27
Add IPTS support for 5.5+

StollD:
I am reasonably convinced that the new driver does the stuff that it can do good enough that we can open it up to a bigger testing group by including it into our builds. 

Although I don't think there are any other stylus protocols except for gen4 and gen5+, adding this to the prebuilt kernel will make it easier to discover those variances since more users will use and test it. Regarding singletouch, it is pretty much guaranteed that this data will be the same on all devices.

The worst case scenario is that it still won't work correctly for some people (no difference to 5.4). But it might allow others (me, lol), to finally move on from 5.3.y, which is ridicioulusly out of date.

**Note:** I don't really remember what the state of the driver on gen7 was, but I included the code to make it probe there. If this turns out to be a problem we could remove it, and only target gen4-6 with it.


archseer:
Fully agree! I'll give it a spin on gen7 next weekend.

archseer:
This is without multitouch support, right?

StollD:
Yes, this is without multitouch support, and most likely without stylus support for gen7 as well.

PR #36, opened 2020-02-27
v5.5: mwifiex: disable ps_mode by default

kitakar5525:
(This PR should also apply cleanly on 4.19 and 5.4)

It's known that ps_mode (power_save) causes connection instability. The change https://github.com/linux-surface/kernel/commit/264427bb5816c913a1abab9de59639eec299d972#diff-b645af90396992a038c24ef0d70bd262 reverted lines that forcibly overrides ps_mode to false.

And according to some reports, it seems that disabling ps_mode via config file is not so reliable.

So, this PR change back the ps_mode to disabled by default. Also, there may be no reason to keep it enabled by default if it's broken.

- What do you think, everyone? Should we really modify the driver to disable ps_mode by default by kernel side again?

---

I first tried to stop enabling ps_mode on initialization ("mwifiex: sta_cmd: do not enable ps_mode by default"). But NetworkManager still enabled ps_mode.

So, I also introduced a module parameter called `allow_ps_mode` and disallows enabling ps_mode by default. ("mwifiex: cfg80211: add allow_ps_mode module parameter")

The difference to the previously reverted lines (https://github.com/linux-surface/kernel/commit/264427bb5816c913a1abab9de59639eec299d972#diff-b645af90396992a038c24ef0d70bd262) is that we can still re-enable the ps_mode via module parameter just like other wifi drivers:
- https://github.com/torvalds/linux/blob/ca7e1fd1026c5af6a533b4b5447e1d2f153e28f2/drivers/net/wireless/intel/iwlwifi/iwl-drv.c#L1866-L1872 (`iwlwifi_mod_params.power_save`, iwlwifi)
- https://github.com/torvalds/linux/blob/a2d79c7174aeb43b13020dd53d85a7aefdd9f3e5/drivers/net/wireless/ath/ath9k/init.c#L67-L69 (`ath9k_ps_enable`, ath9k)

Also, note here that those two drivers also seem to disable power_save by default.

I also added a commit to print a message when the ps_mode state is going to be changed. Mainly for printing a warning to the user when ps_mode will get enabled. ("mwifiex: cfg80211: print message when changing ps_mode")

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/cfg80211.c
+++ b/drivers/net/wireless/marvell/mwifiex/cfg80211.c
@@ -457,6 +457,14 @@ mwifiex_cfg80211_set_power_mgmt(struct wiphy *wiphy,
 		return -1;
 	}
 
+	if (ps_mode)
+		dev_warn(priv->adapter->dev,
+			    "WARN: Request to enable ps_mode received. Enabling it. "
+			    "Disable it if you encounter connection instability.\n");
+	else
+		dev_info(priv->adapter->dev,
+			    "Request to disable ps_mode received. Disabling it.\n");
+
 	return mwifiex_drv_set_power(priv, &ps_mode);
 }
 
--- a/drivers/net/wireless/marvell/mwifiex/cfg80211.c
+++ b/drivers/net/wireless/marvell/mwifiex/cfg80211.c
@@ -25,6 +25,11 @@
 static char *reg_alpha2;
 module_param(reg_alpha2, charp, 0);
 
+static bool allow_ps_mode;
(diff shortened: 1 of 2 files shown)


qzed:
> * What do you think, everyone? Should we really modify the driver to disable ps_mode by default by kernel side again?

At the moment I don't see any alternatives. Given that NetworkManager (and probably other tools, too) like do do their own thing regarding power management, your solution seems like the best option to me for now (providing an option to re-enable power-save manually). Normally I don't really like these kinds of hacks (straight-on force disabling stuff), but it seems to me that this may be the only sane option to guarantee that power-save is (and stays) off.

PR #42, opened 2020-04-06
Backport simplified feedback sending and remove no_feedback

StollD:
@kitakar5525 and I figured out that the feedback sending I implemented for the new IPTS driver doesn't appear to cause the crashes on gen4 devices that lead to no_feedback being implemented. I backported it to this driver, and it works here as well: No crashes on a device that required no_feedback before.

The difference between the old implementation and this one is, that the old one got the feedback data that it sent through the GuC. We have no idea what data is sent, and how the ME reacts to it. Because the GuC is not useable on the new driver, it will send a buffer that contains nothing but the transaction / buffer ID. 

This causes an "Invalid Parameters" error from the ME, but touch data continues to be written into the buffers (as opposed to not sending feedback at all). The "Invalid Parameters" error has been silenced on the feedback command for that reason.

The other commits remove the now obsolete code for device-specific quirks. If we ever need quirks again, we can just revert the last commit and then add whatever quirk is required to the base code.

Note: This has only been tested on one device to verify that it fixes the issue with no_feedback. I haven't tested it on any other device, so testing is welcome!

### DIFF
--- a/drivers/misc/ipts/companion.c
+++ b/drivers/misc/ipts/companion.c
@@ -209,22 +209,3 @@ config_fallback:
 	return ret;
 
 }
-
-unsigned int ipts_get_quirks(void)
-{
-	unsigned int ret;
-
-	// Make sure that access to the companion is synchronized
-	mutex_lock(&ipts_companion_lock);
-
-	// If the companion is ignored, or doesn't exist, assume that
-	// the device doesn't have any quirks
-	if (ipts_modparams.ignore_companion || ipts_companion == NULL)
-		ret = IPTS_QUIRK_NONE;
-	else
-		ret = ipts_companion->get_quirks(ipts_companion);
-
-	mutex_unlock(&ipts_companion_lock);
-
-	return ret;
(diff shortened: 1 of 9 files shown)


kitakar5525:
Thanks, this PR works on SB1 !

qzed:
Tested on SB2. Works as expected.

kitakar5525:
Just leaving a comment that I noticed comment for `MSHW0101` and `MSHW0102` are reversed on removed quirk implementation.
"Surface Pro 5 / 6" does use `MSHW0102` and "Surface Book 2 15"" does use `MSHW0101`
https://github.com/linux-surface/kernel/pull/42/commits/eeda2609a689171ac15011ae59ca87bafe3395e1#diff-a7537ec93e04f8161acfbaea24a84a0fL52-L62
(It seems that it became reversed when adding quirk implementation (7909b87), and the newly added code on this PR has correct comments.)

It needs to be taken care of when we ever need to revert the last commit.

PR #44, opened 2020-04-23
mwifiex: split large commit into three commits

kitakar5525:
(The first revert commit does not apply cleanly on v4.19 branch. So, I'll make the same change to v5.4 and v4.19 branch once this PR gets approved. Or let me know if you guys want to do it by yourself.)

This PR
1) reverts the large commit 456cdc416f6e ("wireless/mwifiex: Fix S0ix / suspend")
2) split the commit into three small commits

So, no functional changes are intended. Just split the commit to describe what each part does.

There are small changes from the reverted commit.
1) added a description to the top of suspend()/resume() functions (starting with 'XXX:')
2) used `pci_upstream_bridge()` instead of `bus->self->` and placed `bridge_d3` thing to the top of probe()
3) removed description regarding auto_ds from the top of the function
4) not changing the location of `int ret;`; the previous commit changed the location (I feel this is rather cleaner, but I followed the upstream code)
5) added a comment where the removed code existed
6) added commit message to each commit

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -149,38 +149,35 @@ static bool mwifiex_pcie_ok_to_access_hw(struct mwifiex_adapter *adapter)
  */
 static int mwifiex_pcie_suspend(struct device *dev)
 {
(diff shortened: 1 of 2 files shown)


qzed:
Thanks!

PR #45, opened 2020-04-23
mwifiex: sta_cmd: add comment for not enabling ps_mode by default

kitakar5525:
(once this PR gets approved, I'll make the same change to v4.19 and v5.4 branches or let me know if you guys want to do it by yourself.)

The commit 541813622549 ("mwifiex: sta_cmd: do not enable ps_mode by
default") stopped enabling ps_mode by default by just removing the
relevant code. But it's better to leave a comment there.

This commit adds a comment where the related code existed.

Signed-off-by: Tsuchiya Yuto (kitakar5525) <kitakar@gmail.com>


qzed:
Thanks! Feel free to add this to 4.19. I think 5.4 needs a general update if we consider it for LTS builds and is missing some stuff, so you don't have to add it there (if you still want to, you can of course).

PR #51, opened 2020-05-15
v4.19: ioremap_uc: revert commits that were upstreamed now

kitakar5525:
(The PR title may be misleading due to my English...)
I reverted two commits that were backported by linux-surface, because they're now backported by upstream on v4.19.122.

(Do you think reverting our commits explicitly is needed on devel branch? If not, feel free to close this PR, thanks.)

Note: `ioremap_uc` related patches in v4.19-surface-devel is now only commit de067f5c120969 ("docs: driver-model: add devm_ioremap_uc"), which is not in upstream v4.19 (at least yet?)

### DIFF
--- a/drivers/mfd/intel-lpss.c
+++ b/drivers/mfd/intel-lpss.c
@@ -397,7 +397,7 @@ int intel_lpss_probe(struct device *dev,
 	if (!lpss)
 		return -ENOMEM;
 
-	lpss->priv = devm_ioremap_uc(dev, info->mem->start + LPSS_PRIV_OFFSET,
+	lpss->priv = devm_ioremap(dev, info->mem->start + LPSS_PRIV_OFFSET,
 				  LPSS_PRIV_SIZE);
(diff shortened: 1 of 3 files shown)


kitakar5525:
I think I did a meaningless thing:
- without this PR: If I merge v4.19-surface-devel into newer versions (e.g. tag v4.19.123), it succeeds anyway. Rebasing will also succeeds with `git commit --allow-empty` (our commits now become empty)
- with this PR: If I merge v4.19-surface-devel then rebase with `git commit --allow-empty`, our commits now become empty and those revert commits from this PR actually revert code. So, even harmful.

- merging (current) v4.19-surface into v4.19.123 will also automatically succeeds.

and reverting our commits may be even misleading. Closing this PR.

qzed:
We could rebase to remove those (then empty) commits. I don't think that's necessary yet though, but we should keep that in mind if we rebase 4.19-surface-devel at some point in the future.

PR #58, opened 2020-09-24
mwifiex: add fw reset quirks for Surface devices

kitakar5525:
This PR adds
1. DMI tables of Surface devices (first and second patch)
2. fw reset quirk for Surface gen4+ devices (third patch)
3. fw reset quirk for Surface 3 (fourth patch)

While here, created new files for quirks (`mwifiex/pcie_quirks.c` and `mwifiex/pcie_quirks.h`) because the changes are a little bit too big to add into `pcie.c`.

For Surface gen4+ devices, we can reset fw by putting wifi into D3cold then bring back into D0 (power-cycle).

For Surface 3, we can reset fw by calling _DSM method available in `\_SB.WSID` device (https://github.com/linux-surface/acpidumps/blob/05cba925f3a515f222acb5b3551a032ddde958fe/surface_3/dsdt.dsl#L11947-L12011).

Note that for Surface Pro 3, there is the same `WSID` device exists (https://github.com/linux-surface/acpidumps/blob/05cba925f3a515f222acb5b3551a032ddde958fe/surface_pro_3/dsdt.dsl#L12164-L12216). However, I'm not sure if it will work in the same way as Surface 3. So, I didn't add support for SP3 in this PR. Might look into it later.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -396,6 +396,13 @@ static void mwifiex_pcie_reset_prepare(struct pci_dev *pdev)
 	mwifiex_shutdown_sw(adapter);
 	clear_bit(MWIFIEX_IFACE_WORK_DEVICE_DUMP, &card->work_flags);
 	clear_bit(MWIFIEX_IFACE_WORK_CARD_RESET, &card->work_flags);
+
+	/* For Surface gen4+ devices, we need to put wifi into D3cold right
+	 * before performing FLR
+	 */
+	if (card->quirks & QUIRK_FW_RST_D3COLD)
+		mwifiex_pcie_reset_d3cold_quirk(pdev);
+
 	mwifiex_dbg(adapter, INFO, "%s, successful\n", __func__);
 
 	card->pci_reset_ongoing = true;
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -2833,6 +2833,16 @@ static void mwifiex_pcie_card_reset_work(struct mwifiex_adapter *adapter)
 {
 	struct pcie_service_card *card = adapter->card;
 
+	/* On Surface 3, reset_wsid method removes then re-probes card by
+	 * itself. So, need to place it here and skip performing any other
+	 * reset-related works.
+	 */
+	if (card->quirks & QUIRK_FW_RST_WSID_S3) {
+		mwifiex_pcie_reset_wsid_quirk(card->dev);
+		/* skip performing any other reset-related works */
+		return;
+	}
+
 	/* We can't afford to wait here; remove() might be waiting on us. If we
 	 * can't grab the device lock, maybe we'll get another chance later.
 	 */
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -27,6 +27,7 @@
 #include "wmm.h"
 #include "11n.h"
 #include "pcie.h"
+#include "pcie_quirks.h"
 
 #define PCIE_VERSION	"1.0"
 #define DRV_NAME        "Marvell mwifiex PCIe"
@@ -276,6 +277,9 @@ static int mwifiex_pcie_probe(struct pci_dev *pdev,
 		if (ret)
 			return ret;
 	}
+
+	/* check quirks */
+	mwifiex_initialize_quirks(card);
 
 	if (mwifiex_add_card(card, &card->fw_done, &pcie_ops,
 			     MWIFIEX_PCIE, &pdev->dev)) {
--- a/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
@@ -1,0 +1,114 @@
+// SPDX-License-Identifier: GPL-2.0
+/*
+ * File for PCIe quirks.
+ */
+
+/* The low-level PCI operations will be performed in this file. Therefore,
+ * let's use dev_*() instead of mwifiex_dbg() here to avoid troubles (e.g.
+ * to avoid using mwifiex_adapter struct before init or wifi is powered
+ * down, or causes NULL ptr deref).
+ */
+
+#include <linux/dmi.h>
+
+#include "pcie_quirks.h"
+
+/* quirk table based on DMI matching */
+static const struct dmi_system_id mwifiex_quirk_table[] = {
+	{
+		.ident = "Surface Pro 4",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 4"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Pro 5",
+		.matches = {
+			/* match for SKU here due to generic product name "Surface Pro" */
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_SKU, "Surface_Pro_1796"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Pro 5 (LTE)",
+		.matches = {
+			/* match for SKU here due to generic product name "Surface Pro" */
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_SKU, "Surface_Pro_1807"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Pro 6",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 6"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Book 1",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Book"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Book 2",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Book 2"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Laptop 1",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Laptop"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Laptop 2",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Laptop 2"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface 3",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface 3"),
+		},
+		.driver_data = 0,
+	},
+	{
+		.ident = "Surface Pro 3",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 3"),
+		},
+		.driver_data = 0,
+	},
+	{}
+};
+
+void mwifiex_initialize_quirks(struct pcie_service_card *card)
+{
+	struct pci_dev *pdev = card->dev;
+	const struct dmi_system_id *dmi_id;
+
+	dmi_id = dmi_first_match(mwifiex_quirk_table);
+	if (dmi_id)
+		card->quirks = (uintptr_t)dmi_id->driver_data;
+
+	if (!card->quirks)
+		dev_info(&pdev->dev, "no quirks enabled\n");
+}
--- a/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
@@ -21,7 +21,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 4"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Pro 5",
@@ -30,7 +30,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_SKU, "Surface_Pro_1796"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Pro 5 (LTE)",
@@ -39,7 +39,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_SKU, "Surface_Pro_1807"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Pro 6",
@@ -47,7 +47,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 6"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Book 1",
@@ -55,7 +55,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Book"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Book 2",
@@ -63,7 +63,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Book 2"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Laptop 1",
@@ -71,7 +71,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Laptop"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface Laptop 2",
@@ -79,7 +79,7 @@ static const struct dmi_system_id mwifiex_quirk_table[] = {
 			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
 			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Laptop 2"),
 		},
-		.driver_data = 0,
+		.driver_data = (void *)QUIRK_FW_RST_D3COLD,
 	},
 	{
 		.ident = "Surface 3",
@@ -111,4 +111,61 @@ void mwifiex_initialize_quirks(struct pcie_service_card *card)
 
 	if (!card->quirks)
 		dev_info(&pdev->dev, "no quirks enabled\n");
+	if (card->quirks & QUIRK_FW_RST_D3COLD)
+		dev_info(&pdev->dev, "quirk reset_d3cold enabled\n");
+}
+
+static void mwifiex_pcie_set_power_d3cold(struct pci_dev *pdev)
+{
+	dev_info(&pdev->dev, "putting into D3cold...\n");
+
+	pci_save_state(pdev);
+	if (pci_is_enabled(pdev))
+		pci_disable_device(pdev);
+	pci_set_power_state(pdev, PCI_D3cold);
+}
+
+static int mwifiex_pcie_set_power_d0(struct pci_dev *pdev)
+{
+	int ret;
+
+	dev_info(&pdev->dev, "putting into D0...\n");
+
+	pci_set_power_state(pdev, PCI_D0);
+	ret = pci_enable_device(pdev);
+	if (ret) {
+		dev_err(&pdev->dev, "pci_enable_device failed\n");
+		return ret;
+	}
+	pci_restore_state(pdev);
+
+	return 0;
+}
+
+int mwifiex_pcie_reset_d3cold_quirk(struct pci_dev *pdev)
+{
+	struct pci_dev *parent_pdev = pci_upstream_bridge(pdev);
+	int ret;
+
+	/* Power-cycle (put into D3cold then D0) */
+	dev_info(&pdev->dev, "Using reset_d3cold quirk to perform FW reset\n");
+
+	/* We need to perform power-cycle also for bridge of wifi because
+	 * on some devices (e.g. Surface Book 1), the OS for some reasons
+	 * can't know the real power state of the bridge.
+	 * When tried to power-cycle only wifi, the reset failed with the
+	 * following dmesg log:
+	 * "Cannot transition to power state D0 for parent in D3hot".
+	 */
+	mwifiex_pcie_set_power_d3cold(pdev);
+	mwifiex_pcie_set_power_d3cold(parent_pdev);
+
+	ret = mwifiex_pcie_set_power_d0(parent_pdev);
+	if (ret)
+		return ret;
+	ret = mwifiex_pcie_set_power_d0(pdev);
+	if (ret)
+		return ret;
+
+	return 0;
 }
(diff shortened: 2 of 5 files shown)


kitakar5525:
You can trigger the card reset manually via sysfs entry:

    $ echo 1 | sudo tee /sys/kernel/debug/mwifiex/mlan0/reset

kitakar5525 (commented):
`mwifiex: pcie: add reset_wsid quirk for Surface 3`

Forgot to mention that why this quirk is required.

```
To reset mwifiex on Surface 3, it seems that calling the _DSM method
exists in \_SB.WSID [1] device is required.

[1] https://github.com/linux-surface/acpidumps/blob/05cba925f3a515f222acb5b3551a032ddde958fe/surface_3/dsdt.dsl#L11947-L12011
```

kitakar5525 (commented):
`mwifiex: pcie: add reset_wsid quirk for Surface 3`

Hmm, some indentations look weird. I'll fix them later.

kitakar5525:
Addressed the following checkpatch warnings:
`mwifiex: pcie: add Surface devices DMI table for quirks`
```
WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
#33: FILE: drivers/net/wireless/marvell/mwifiex/pcie_quirks.h:1:
+/*
```

`mwifiex: pcie: add reset_d3cold quirk for Surface gen4+ devices`
```
WARNING: Block comments use a trailing */ on a separate line
#53: FILE: drivers/net/wireless/marvell/mwifiex/pcie.c:398:
+	 * before performing FLR */

WARNING: Missing or malformed SPDX-License-Identifier tag in line 1
#66: FILE: drivers/net/wireless/marvell/mwifiex/pcie_quirks.c:1:
+/*

WARNING: Block comments use a trailing */ on a separate line
#109: FILE: drivers/net/wireless/marvell/mwifiex/pcie_quirks.c:44:
+	 * "Cannot transition to power state D0 for parent in D3hot". */
```

`mwifiex: pcie: add reset_wsid quirk for Surface 3`
```
WARNING: Block comments use a trailing */ on a separate line
#39: FILE: drivers/net/wireless/marvell/mwifiex/pcie.c:2830:
+	 * reset-related works. */
```

Ignoring `WARNING: networking block comments don't use an empty /* line, use /* Comment...` at top of files in accordance with the other upstream files:
```
WARNING: networking block comments don't use an empty /* line, use /* Comment...
#35: FILE: drivers/net/wireless/marvell/mwifiex/pcie_quirks.h:3:
+/*
+ * Header file for PCIe quirks.
```

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
```diff
@@ -0,0 +1,121 @@
...
+ * File for PCIe quirks.
+ */
+
+#include <linux/acpi.h>
+#include <linux/pci.h>
+
+#include "pcie_quirks.h"
+
+/* For reset_wsid quirk */
+#define ACPI_WSID_PATH		\\_SB.WSID"
```
qzed, line 12: There's an open-quote (`"`) missing.

drivers/net/wireless/marvell/mwifiex/pcie.c
```diff
@@ -392,6 +393,13 @@ static void mwifiex_pcie_reset_prepare(struct pci_dev *pdev)
 	mwifiex_shutdown_sw(adapter);
 	clear_bit(MWIFIEX_IFACE_WORK_DEVICE_DUMP, &card->work_flags);
 	clear_bit(MWIFIEX_IFACE_WORK_CARD_RESET, &card->work_flags);
+
+	/* For Surface gen4+ devices, we need to put wifi into D3cold right
+	 * before performing FLR
+	 */
+	if (dmi_check_system(surface_gen4p_dmi_table))
```
qzed, line 400: Quirks are usually implemented via bitflags, here that would be on the `card` I think, so the DMI matching is done once when the driver is loaded and based on that, bitflags are set. Then check those bitflags here and below instead of DMI matching.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
```diff
@@ -0,0 +1,104 @@
...
+static const struct dmi_system_id surfacepro3_dmi_table[] = {
+	{
+		.ident = "Surface Pro 3",
+		.matches = {
+			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
+			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 3"),
+		},
+	},
+	{}
+};
```
qzed, line 101: With bitflags you could combine all of those entries into one table, then match them at the beginning and set the bitflags of the card correspondingly. You can use the `driver_data` member to store a reference to the quirks.

Also it's generally a bad idea to have global `static const` variables in your header file. This means that they will be included in every compilation unit, so they will be duplicated. The usual way to set up global variables is via, in your `.h` file
```c
extern const struct dmi_system_id surface3_dmi_table[];
```
and in one of your `.c` files:
```c
const struct dmi_system_id surface3_dmi_table[] = {
    // ...
};
```

In this case though, it's probably better if you make the quirks fully private to the `.c` file and add a function like `mwifiex_initialize_quirks()` or something to set up the bitflags on `card`.

qzed (changes requested):

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Ah, I've messed up when rebasing. Resolved by commit https://github.com/linux-surface/kernel/pull/58/commits/7610032daf1c0ec557e499a72c53f0324cd9e8d4

drivers/net/wireless/marvell/mwifiex/pcie.c
kitakar5525: Got it. Addressed with commit https://github.com/linux-surface/kernel/pull/58/commits/206fa3cd1d885b4bb5d72b8b518a2b90d580d548

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
kitakar5525: > With bitflags you could combine all of those entries into one table, then match them at the beginning and set the bitflags of the card correspondingly. You can use the `driver_data` member to store a reference to the quirks.

OK, I've switched to using `driver_data` and combined all the entries into one table.

> Also it's generally a bad idea to have global `static const` variables in your header file. This means that they will be included in every compilation unit, so they will be duplicated. The usual way to set up global variables is via, in your `.h` file
> 
> ```c
> extern const struct dmi_system_id surface3_dmi_table[];
> ```
> 
> and in one of your `.c` files:
> 
> ```c
> const struct dmi_system_id surface3_dmi_table[] = {
>     // ...
> };
> ```

Thanks. For record, indeed I see only one place in upstream tree where DMI table exists in `.h` file (https://github.com/torvalds/linux/blob/master/drivers/input/serio/i8042-x86ia64io.h). It uses `__initconst`.

> In this case though, it's probably better if you make the quirks fully private to the `.c` file and add a function like `mwifiex_initialize_quirks()` or something to set up the bitflags on `card`.

OK, I've placed DMI table into `pcie_quirks.c` instead. The quirks will be set up by `mwifiex_initialize_quirks()`

Addressed by commit https://github.com/linux-surface/kernel/pull/58/commits/206fa3cd1d885b4bb5d72b8b518a2b90d580d548

kitakar5525:
Thanks for the review! I've pushed four new patches.

The first two patches addressed the review.
The rest is for simplification/upstreaming.

If they look OK, I'll squash the patches.

drivers/net/wireless/marvell/mwifiex/pcie.c
```diff
@@ -284,6 +284,9 @@ static int mwifiex_pcie_probe(struct pci_dev *pdev,
 		return -1;
 	}
 
+	/* check quirks */
+	mwifiex_initialize_quirks(card);
```
qzed, line 288: I think this should be done before calling `mwifiex_add_card()`, so that it's set before it can be accessed somewhere else.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
```diff
@@ -3,102 +3,17 @@
...
-		.ident = "Surface Pro 3",
-		.matches = {
-			DMI_EXACT_MATCH(DMI_SYS_VENDOR, "Microsoft Corporation"),
-			DMI_EXACT_MATCH(DMI_PRODUCT_NAME, "Surface Pro 3"),
-		},
-	},
-	{}
-};
+/* Quirks */
+#define QUIRK_NONE		0
```
qzed, line 9: I think you can drop this. If there isn't any bit set it should be implied that there's no quirk required.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
```diff
@@ -0,0 +1,248 @@
...
+	mwifiex_pcie_set_power_d0(pdev);
+
+	return 0;
+}
+
+int mwifiex_pcie_reset_wsid_quirk(struct pci_dev *pdev)
+{
+	struct pcie_service_card *card = pci_get_drvdata(pdev);
+	struct mwifiex_adapter *adapter = card->adapter;
+	char *path = ACPI_WSID_PATH;
```
qzed, line 196: This should probably be `const`. You could also drop this and use the macro directly. For printing it, you can use
```c
dev_err(&pdev->dev, "No ACPI handle for path "ACPI_WSID_PATH "\n" );
```

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
```diff
@@ -0,0 +1,248 @@
...
+	static guid_t wsid_dsm_guid;
```
qzed, line 197: This should probably not be static. As mentioned in the code below, you can also just use `GUID_INIT()` and make it a `static const` global.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
```diff
@@ -0,0 +1,248 @@
...
+
+	/* card will be removed immediately after this call on Surface 3 */
+	mwifiex_dbg(adapter, MSG, "turning wifi off...\n");
+	obj = acpi_evaluate_dsm(handle, &wsid_dsm_guid,
+				WSID_REV, WSID_FUNC_WIFI_PWR_OFF,
+				NULL);
+	if (!obj) {
+		dev_err(&pdev->dev,
+			"device _DSM execution failed for turning wifi off\n");
+		return -ENODEV;
```
qzed, line 233: Not sure if `-ENODEV` is the best error to return here. Maybe `-EFAULT` is better, but I'm not really sure. Same with the evaluation below.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
```diff
@@ -0,0 +1,248 @@
...
+	char *path = ACPI_WSID_PATH;
+	static guid_t wsid_dsm_guid;
+	acpi_handle handle;
+	union acpi_object *obj;
+	acpi_status status;
+
+	mwifiex_dbg(adapter, MSG,
+		    "Using reset_wsid quirk to perform FW reset\n");
+
+	guid_parse(WSID_DSM_UUID, &wsid_dsm_guid);
```
qzed, line 205: I think the preferred way is to use `GUID_INIT()` instead of defining the GUID as string, but this will work too. Not sure if upstream will complain about that.

qzed (changes requested):
Looks mostly good, a couple of nitpicks.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
kitakar5525: I thought I can use `QUIRK_NONE` when adding quirk implementation like `.driver_data = (void *)QUIRK_NONE,` (and the actual quirks will come in later commits).

That said, I can't find the real usage in upstream tree yet. In this case, using `0` directly is preferred? (`.driver_data = 0,`). I'll keep the `QUIRK_NONE` for now.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Regarding the constifying, `acpi_get_handle()` expects it to be `expected ‘acpi_string’ {aka ‘char *’}`. So, I'll keep it non-const.
```
  CC [M]  drivers/net/wireless/marvell/mwifiex/pcie_quirks.o
drivers/net/wireless/marvell/mwifiex/pcie_quirks.c: In function ‘mwifiex_pcie_reset_wsid_quirk’:
drivers/net/wireless/marvell/mwifiex/pcie_quirks.c:197:33: warning: passing argument 2 of ‘acpi_get_handle’ discards ‘const’ qualifier from pointer target type [-Wdiscarded-qualifiers]
  197 |  status = acpi_get_handle(NULL, path, &handle);
      |                                 ^~~~
In file included from ./include/acpi/acpi.h:31,
                 from ./include/linux/acpi.h:22,
                 from drivers/net/wireless/marvell/mwifiex/pcie_quirks.c:6:
./include/acpi/acpixf.h:527:23: note: expected ‘acpi_string’ {aka ‘char *’} but argument is of type ‘const char *’
  527 |           acpi_string pathname,
      |           ~~~~~~~~~~~~^~~~~~~~
./include/acpi/acpixf.h:61:2: note: in definition of macro ‘ACPI_EXTERNAL_RETURN_STATUS’
   61 |  prototype;
      |  ^~~~~~~~~
```

With regard to using macro directly, I didn't know I can use it directly 😲 It seems that
```c
dev_err(&pdev->dev, "No ACPI handle for path %s\n", ACPI_WSID_PATH );
```
also works.

That said, I can't find the usage like this in upstream tree. Do you have the example? I'll keep current way for now.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Hmm, upstream uses various return values for `acpi_evaluate_dsm()` failure:
- return -ENXIO;
- return -ENODEV;
- return -ENOTSUPP;
- return -EINVAL;
- return -EIO;

In this case, it just means that execution has failed. So, as you mention, `-EFAULT` may be better. I'll change this in the next rebase.

By the way, it seems that I've forgotten using `ACPI_FREE(obj);`, I'll add in the next rebase.

kitakar5525:
Squashed patches with addressing the review feedbacks:
- moved `mwifiex_initialize_quirks()` before `mwifiex_add_card()`
- switched to using `GUID_INIT()`
- changed `acpi_evaluate_dsm()` failure return value to `-EFAULT`
- added `ACPI_FREE(obj);` after `acpi_evaluate_dsm()` success

Other changes:
- do not use `mwifiex_dbg()` to avoid troubles in pcie_quirks.c
  (e.g. to avoid using mwifiex_adapter struct before init or wifi is powered down, or causes NULL ptr deref).
  Added comment regarding this to the top of pcie_quirks.c:
  https://github.com/linux-surface/kernel/blob/5dcea7f78ec512698da3b6ba1d75c346147bc438/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c#L21-L25

drivers/net/wireless/marvell/mwifiex/pcie.c
kitakar5525: Moved `mwifiex_initialize_quirks()` before `mwifiex_add_card()` by the new rebase.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Ah I forgot to remove `static`. Anyway, switched to using `GUID_INIT()` by the new rebase.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: - changed `acpi_evaluate_dsm()` failure return value to `-EFAULT`
- added `ACPI_FREE(obj);` after `acpi_evaluate_dsm()` success

Addressed by the new rebase.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Switched to using GUID_INIT() by the new rebase.

kitakar5525:
I've rebased again.

Changes:
- Fixed changes squashed into wrong patch (the print `dev_info(&pdev->dev, "no quirks applied\n");` went into second patch by mistake, now in first patch)
- Moved comment for (`do not use mwifiex_dbg() to avoid troubles in pcie_quirks.c`) higher (now above `#include`):
  https://github.com/linux-surface/kernel/blob/a90d243a0e24de888de0602d9c71db21deaa925e/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c#L1-L15
- Use result of `mwifiex_pcie_set_power_d0()`
  https://github.com/linux-surface/kernel/blob/a90d243a0e24de888de0602d9c71db21deaa925e/drivers/net/wireless/marvell/mwifiex/pcie_quirks.c#L186-L191

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
qzed: I think `= 0` directly is fine here.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
qzed: Static string concatenation is, for example, often used in `pr_fmt` definitions (e.g. [here](https://elixir.bootlin.com/linux/v5.8.12/source/drivers/cpufreq/pmac32-cpufreq.c#L12), where you have `KBUILD_MODNAME` and `fmt` as static strings, and the `": "` as separator).

Basically all that does is concatenate two static strings. You could also write `char *str = "abc" "def"` which then would get compiled to `char *str = "abcdef"`. The difference between
```c
dev_err(&pdev->dev, "No ACPI handle for path " ACPI_WSID_PATH "\n");
```
and
```c
dev_err(&pdev->dev, "No ACPI handle for path %s\n", ACPI_WSID_PATH);
```

is that the first will actually concatenate the strings at compile time, whereas the second will insert the string dynamically at runtime.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
qzed: I think `-EIO` would be fine too. `-ENODEV` and `-ENXIO` (to me) seem more like "the device or expected method is not available", which you already checked previously. `-ENOTSUPP` is similar, just indicates that that might be something that's not expected to be there. `-EINVAL` is more for input dependent errors, i.e. invalid input to the function calling the _DSM.

In our case, I think it makes more sense to use an error value that signals "the function is there, we tried to execute it, but something during execution failed".

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
qzed: I think `= 0` is fine here.

Although that's really only my own preference. I think that bit-flags imply that zero is none, the same way you don't do `FLAG_NONE | FLAG_THING_ONE | FLAG_THING_TWO` and just `FLAG_THING_ONE | FLAG_THING_TWO`.

Also since it's static data, you could also drop the `.driver_data = 0` entirely as the default initialization covers that and will initialize it to zero.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
qzed: And right, the `acpi_get_handle` not taking a `const char *` is a bit annoying...

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
kitakar5525: > I think `= 0` is fine here.
> 
> Although that's really only my own preference. I think that bit-flags imply that zero is none, the same way you don't do `FLAG_NONE | FLAG_THING_ONE | FLAG_THING_TWO` and just `FLAG_THING_ONE | FLAG_THING_TWO`.

OK.

> Also since it's static data, you could also drop the `.driver_data = 0` entirely as the default initialization covers that and will initialize it to zero.

The first patch is a implementation, I want to preserve the skeleton as much as possible. Also I see a few usage of `.driver_data = 0` in upstream. So, I'll go this way (use `.driver_data = 0` and drop `QUIRK_NONE`).

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Looks like the upstream handles this in a few ways:

Using macro directly and using concatenating on print:
- https://github.com/torvalds/linux/blob/master/drivers/hwmon/asus_atk0110.c#L1213-L1219
```c
	/* RVLT: read voltage */
	status = acpi_get_handle(data->atk_handle, METHOD_OLD_READ_VLT, &ret);
	if (ACPI_SUCCESS(status))
		data->rvlt_handle = ret;
	else
		dev_dbg(dev, "method " METHOD_OLD_READ_VLT " not found: %s\n",
				acpi_format_exception(status));
```

Casting the const variable to `(char *)`
- https://github.com/torvalds/linux/blob/master/drivers/platform/x86/wmi.c#L162
- https://github.com/torvalds/linux/blob/master/drivers/platform/x86/asus-laptop.c#L396-L410
- https://github.com/torvalds/linux/blob/master/drivers/platform/x86/eeepc-laptop.c#L247-L252
```c
	status = acpi_get_handle(eeepc->handle, (char *)method,
				 handle);
	if (status != AE_OK) {
		pr_warn("Error finding %s\n", method);
		return -ENODEV;
	}
```

Casting to `(acpi_string)`
- https://github.com/torvalds/linux/blob/master/drivers/platform/x86/lg-laptop.c#L97
- https://github.com/torvalds/linux/blob/master/drivers/net/wireless/ath/ath10k/mac.c#L9001
```c
	status = acpi_get_handle(root_handle, (acpi_string)WRD_METHOD, &handle);
```

Using macro directly and using format on print:
- https://github.com/torvalds/linux/blob/master/drivers/acpi/acpica/dbtest.c#L187-L194
```c
		status =
		    acpi_get_handle(NULL, ACPI_DB_READ_METHOD, &read_handle);
		if (ACPI_FAILURE(status)) {
			acpi_os_printf
			    ("Could not obtain handle for debug method %s\n",
			     ACPI_DB_READ_METHOD);
			return;
		}
```

The last example is from `acpica` directory. So, I'll use the style and drop the `path` variable.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: - https://github.com/torvalds/linux/blob/02de58b24d2e1b2cf947d57205bd2221d897193c/include/uapi/asm-generic/errno-base.h#L9
```c
#define	EIO		 5	/* I/O error */
```

- https://github.com/torvalds/linux/blob/02de58b24d2e1b2cf947d57205bd2221d897193c/include/uapi/asm-generic/errno-base.h#L18
```c
#define	EFAULT		14	/* Bad address */
```

Honestly, I'm not sure which is more appropriate in this case. So, I'll again change the return value to `-EIO` this time considering that I see many usage in upsteam.

kitakar5525:
Rebased with addressing review feedbacks:
- dropped `QUIRK_NONE` and used `.driver_data = 0` for the skeleton
- dropped `path` variable in `mwifiex_pcie_reset_wsid_quirk()`
- changed return value of `acpi_evaluate_dsm()` failure to `-EIO`

Other changes:
- added _DSM check for also wifi power on function
  There was only check for wifi power off function before, so also add for `on` accordingly
- change wording `no quirks applied` -> `no quirks enabled` in `mwifiex_initialize_quirks()` in accordance with the other print there
- just use "NOT operator" instead of `==` in `mwifiex_initialize_quirks()`
  `card->quirks == 0` -> `!card->quirks`
- moved patch `mwifiex: pcie: (OEMB) add quirk for Surface 3 with broken DMI table` to last so that the first three patches can be sent upstream easily

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
kitakar5525: Addressed with the new rebase.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: Addressed with the new rebase.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.c
kitakar5525: changed return value of `acpi_evaluate_dsm()` failure to `-EIO`

qzed (approved):

qzed:
Okay, looks good to me. Feel free to merge when you're ready.

kitakar5525:
Thanks!

There are still issues remaining:
1. For the reset_wsid quirk (for S3 and SP3), I think that creating a new device driver for the `WSID` device is better, then export the reset function, and call it from mwifiex. As the comments in code states, on S3, calling _DSM methods immediately remove/re-probe mwifiex. So, resetting the wifi externally is a better idea. (Also in this way, hopefully supporting SP3 may become easier.)

2. After reset, ASPM settings don't get restored completely.
`sudo lspci -nnvvv` diff before/after fw reset on SB1:
```diff
#
# 03:00.0 Ethernet controller [0200]: Marvell Technology Group Ltd. 88W8897 [AVASTAR] 802.11ac Wireless [11ab:2b38]
#
@@ -574,9 +574,9 @@
        Capabilities: [168 v1] L1 PM Substates
                L1SubCap: PCI-PM_L1.2+ PCI-PM_L1.1+ ASPM_L1.2+ ASPM_L1.1+ L1_PM_Substates+
                          PortCommonModeRestoreTime=70us PortTPowerOnTime=10us
-               L1SubCtl1: PCI-PM_L1.2+ PCI-PM_L1.1+ ASPM_L1.2+ ASPM_L1.1+
-                          T_CommonMode=0us LTR1.2_Threshold=163840ns
-               L1SubCtl2: T_PwrOn=44us
+               L1SubCtl1: PCI-PM_L1.2- PCI-PM_L1.1- ASPM_L1.2- ASPM_L1.1-
+                          T_CommonMode=0us LTR1.2_Threshold=0ns
+               L1SubCtl2: T_PwrOn=10us
        Kernel driver in use: mwifiex_pcie
        Kernel modules: mwifiex_pcie

#
# no changes on root port of wifi regarding ASPM
#
```
All of the L1 substates are disabled after fw reset. LTR value is also changed.

`sudo lspci -nnvvv` diff before/after fw reset on S3:
```diff
#
# root port of wifi:
# 00:1c.0 PCI bridge [0604]: Intel Corporation Atom/Celeron/Pentium Processor x5-E8000/J3xxx/N3xxx Series PCI Express Port #1 [8086:22c8] (rev 20) (prog-if 00 [Normal decode])
#
@@ -103,7 +103,7 @@
                DevSta: CorrErr+ NonFatalErr- FatalErr- UnsupReq- AuxPwr+ TransPend-
                LnkCap: Port #1, Speed 5GT/s, Width x1, ASPM L0s L1, Exit Latency L0s <512ns, L1 <16us
                        ClockPM- Surprise- LLActRep+ BwNot+ ASPMOptComp-
-               LnkCtl: ASPM L0s L1 Enabled; RCB 64 bytes, Disabled- CommClk+
+               LnkCtl: ASPM Disabled; RCB 64 bytes, Disabled- CommClk+
                        ExtSynch- ClockPM- AutWidDis- BWInt- AutBWInt-
                LnkSta: Speed 2.5GT/s (downgraded), Width x1 (ok)
                        TrErr- Train- SlotClk+ DLActive+ BWMgmt+ ABWMgmt-

#
# no changes on wifi regarding ASPM
#
```
It's worse in this case. The ASPM of wifi root port is disabled after fw reset (thus breaks S0ix).

Anyway, we can improve these later. This PR just works regarding fw reset. So, I'll go this way for now.
I'll merge this later when the other open PR becomes ready.

qzed:
Keep in mind that with an external driver you might need to ensure correct suspend/resume ordering though, which might complicate things. I know too little about the whole situation to say what's better here though-

kitakar5525:
I think what we need to do is just exposing reset function via `EXPORT_SYMBOL_GPL()`. So, I hope there are no issues with suspend/resume ordering, but I may be wrong though.

qzed:
You'd have to make sure that reset can be called both before and after mwifiex has resumed and also executed concurrently. At least if you plan on calling it during resume. Otherwise you'd have to enforce an ordering. You could probably also call it during the `complete` phase, assuming that mwifiex doesn't have a handler for that phase (then it's guaranteed to run after the basic resume steps have been executed). You probably still want to make sure that it can executed concurrently though.

kitakar5525:
Hmm, anyway I'll try something first.

kitakar5525:
Rebased on top of current v5.8-surface-devel due to conflict with https://github.com/linux-surface/kernel/pull/59

katamayko:
friends
i installed debian 12 for the first time this week and i'm having wifi problems on my surface pro 4
I don't understand how to overcome this, can you give me a step-by-step guide if there is a solution

m-doescode:
> friends i installed debian 12 for the first time this week and i'm having wifi problems on my surface pro 4 I don't understand how to overcome this, can you give me a step-by-step guide if there is a solution

Likewise but on Arch

PR #57, opened 2020-09-24
mwifiex: pcie: skip cancel_work_sync() on reset failure path

kitakar5525:
When a reset performed but even that reset failed for some reasons (e.g.
on Surface devices, fw reset requires another quirk), a hang happens on
cancel_work_sync().

    # reset performed after firmware went into bad state
    kernel: mwifiex_pcie 0000:01:00.0: WLAN FW already running! Skip FW dnld
    kernel: mwifiex_pcie 0000:01:00.0: WLAN FW is active
    # but even that reset failed
    kernel: mwifiex_pcie 0000:01:00.0: mwifiex_cmd_timeout_func: Timeout cmd id = 0xfa, act = 0xa000
    kernel: mwifiex_pcie 0000:01:00.0: num_data_h2c_failure = 0
    kernel: mwifiex_pcie 0000:01:00.0: num_cmd_h2c_failure = 0
    kernel: mwifiex_pcie 0000:01:00.0: is_cmd_timedout = 1
    kernel: mwifiex_pcie 0000:01:00.0: num_tx_timeout = 0
    kernel: mwifiex_pcie 0000:01:00.0: last_cmd_index = 2
    kernel: mwifiex_pcie 0000:01:00.0: last_cmd_id: 16 00 a4 00 fa 00 a4 00 7f 00
    kernel: mwifiex_pcie 0000:01:00.0: last_cmd_act: 00 00 00 00 00 a0 00 00 00 00
    kernel: mwifiex_pcie 0000:01:00.0: last_cmd_resp_index = 0
    kernel: mwifiex_pcie 0000:01:00.0: last_cmd_resp_id: 16 80 7f 80 16 80 a4 80 7f 80
    kernel: mwifiex_pcie 0000:01:00.0: last_event_index = 1
    kernel: mwifiex_pcie 0000:01:00.0: last_event: 58 00 58 00 58 00 58 00 58 00
    kernel: mwifiex_pcie 0000:01:00.0: data_sent=0 cmd_sent=1
    kernel: mwifiex_pcie 0000:01:00.0: ps_mode=0 ps_state=0
    kernel: mwifiex_pcie 0000:01:00.0: info: _mwifiex_fw_dpc: unregister device
    # mwifiex_pcie_work hanged
    kernel: INFO: task kworker/0:0:24857 blocked for more than 122 seconds.
    kernel:       Tainted: G        W  OE     5.3.11-arch1-1 #1
[...]

PR #59, opened 2020-09-27
(rework) mwifiex: pcie: skip cancel_work_sync() on reset failure path

kitakar5525:
Hi, this PR is a successor of the previous PR (https://github.com/linux-surface/kernel/pull/57).

This PR
1. reverts previous patch
2. adds updated patch which addressed issues with the previous PR.

The new patch addressed the following:
- Placed `pci_reset_ongoing` variable into `pcie_service_card` (`card->pci_reset_ongoing`) instead of `mwifiex_adapter` (`adapter->pci_reset_ongoing`) (https://github.com/linux-surface/kernel/pull/57#discussion_r495454923)
- Added commit message for why the deadlock happens when reset failed (https://github.com/linux-surface/kernel/pull/57#issuecomment-699493455)

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/main.h
+++ b/drivers/net/wireless/marvell/mwifiex/main.h
@@ -1054,9 +1054,6 @@ struct mwifiex_adapter {
 	void *devdump_data;
 	int devdump_len;
 	struct timer_list devdump_timer;
-
-	/* Indicates a reset is ongoing or not */
-	bool pci_reset_ongoing;
 };
 
 void mwifiex_process_tx_queue(struct mwifiex_adapter *adapter);
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -393,6 +393,8 @@ static void mwifiex_pcie_reset_prepare(struct pci_dev *pdev)
 	clear_bit(MWIFIEX_IFACE_WORK_DEVICE_DUMP, &card->work_flags);
 	clear_bit(MWIFIEX_IFACE_WORK_CARD_RESET, &card->work_flags);
 	mwifiex_dbg(adapter, INFO, "%s, successful\n", __func__);
+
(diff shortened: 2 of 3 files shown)


kitakar5525:
From previous PR (https://github.com/linux-surface/kernel/pull/57#issuecomment-699518964)

> > Also, can I add your above comment to commit message?
> 
> Sure. Btw. you can also add comments to a patch that will not be part of the commit message to the patch by separating the commit message and comment with `---` (the comment then goes between commit message and diff-stat), if you feel that you'd rather include this as comment.

I thought adding the reason why the issue happens into commit message is better.

> > If OK, how can I credit you? (`Co-developed-by`? or just mentioning you in the commit message?)
> 
> Not sure that's really necessary. You developed the patch, so `Co-developed-by` is probably the wrong tag. I just tried to figure out the sequence of why it's going wrong and why this is needed. If you still want to add me, I think the cleanest way is to just mention my first name in the message (or comment) and add my full reference via a `Cc:` tag.

Thanks, I'll `Cc:` you when sending upstream.

kitakar5525:
I'm not sure skipping `cancel_work_sync()` even when reset failed is acceptable or not. I understand `cancel_work_sync()` is there for some reasons. Upstream devs may know anyway.

qzed:
I think it should be fine with reset failure _if_ the device is still being (more or less) usable afterwards and cleanly shut down with a call to `cancel_work_sync()` later (which I kind of assumed). I'd have to take another look at that to see if that's the case. But yeah, ideally we'd move the call to `cancel_work_sync()` to some other place that gives us the necessary guarantees. Problem is I have no clue where, and also what specifically we'd have to guarantee.

qzed (approved):
Thanks! Feel free to merge if you think it's ready.

kitakar5525:
Thanks! since there are no functional changes from the previous PR, I'll merge this PR with the other open PRs later.

kitakar5525:
Corrected wording in commit message in this rebase.

PR #64, opened 2020-10-17
mwifiex: rebase some patches

kitakar5525:
Hi, I slightly cleaned up mwifiex patches.

1. revert some existing patches to improve changes and commit messages
2. added improved patches

`mwifiex: pcie: disable bridge_d3 for Surface gen4+`
Changes from the previous patch:
- Use recently added quirk implementation and apply for only Surface gen4+ devices
- Moved to the bottom of probe() so that quirk implementation can init before use

`mwifiex: add allow_ps_mode module parameter`
Changes from the previous patch:
- Use `-EPERM` instead of `-1` for return value when enabling ps_mode is disallowed by modparam
- changed `allow_ps_mode` modparam permission from `0444` to `0644` because it can be changed after probe as well
- Simplified comment in code
- Simplified print message

`mwifiex: mwifiex: print message when changing ps_mode`
Changes from the previous patch:
- Simplified print messages

`mwifiex: disable ps_mode explicitly by default instead`
Changes from previous patch:
- Squashed `mwifiex: sta_cmd: add comment for not enabling ps_mode by default` and `mwifiex: sta_cmd: do not enable ps_mode by default`
- Disable ps_mode explicitly instead of just removing the code

Removing the code that enables ps_mode might be enough, but there's also
no reason not to explicitly disable it. So, let's disable it explicitly.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/cfg80211.c
+++ b/drivers/net/wireless/marvell/mwifiex/cfg80211.c
@@ -25,6 +25,11 @@
 static char *reg_alpha2;
 module_param(reg_alpha2, charp, 0);
 
+static bool allow_ps_mode;
+module_param(allow_ps_mode, bool, 0644);
+MODULE_PARM_DESC(allow_ps_mode,
+		 "allow WiFi power management to be enabled. (default: disallowed)");
+
 static const struct ieee80211_iface_limit mwifiex_ap_sta_limits[] = {
 	{
 		.max = 3, .types = BIT(NL80211_IFTYPE_STATION) |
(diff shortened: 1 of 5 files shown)


kitakar5525:
I feel that we don't need the patch `mwifiex: pcie: disable bridge_d3 for Surface gen4+` anymore. When I tried on Arch's v5.8 kernel, it survived several suspend. And the bridge_d3 was already set to `false`. But I'm not so sure. For now, let's keep the patch and investigate it later.

qzed:
> I feel that we don't need the patch `mwifiex: pcie: disable bridge_d3 for Surface gen4+` anymore. When I tried on Arch's v5.8 kernel, it survived several suspend. And the bridge_d3 was already set to `false`. But I'm not so sure. For now, let's keep the patch and investigate it later.

Interesting. You could try to add a sysfs file to override the quirks from user-space. That should allow us to test if different things are still needed without having to re-compile things, at least if they're implemented as quirks.

qzed (approved):

PR #66, opened 2020-10-18
surface3-oemb: squash patches into one

kitakar5525:
1. reverted current patches to improve commit message and to squash patches.
2. added squashed one patch

It doesn't make sense to have multiple commits for this issue. So, apply
this workaround treewide in a one patch.

Note that I haven't squashed `mwifiex: pcie: (OEMB) add quirk for Surface 3 with broken DMI table`
because it depends on non-upstream patches.

No changes to the code.


qzed:
I'd normally say we should keep the changes separated (per file changed or per subsystem) if it were intended for upstreaming, but I don't see that happening soon, so I agree that it's probably better to keep all of them in one commit for now.

PR #68, opened 2020-10-22
[5.8] IPTS UAPI v2

StollD:
This implements a new and refined UAPI interface for IPTS that should improve stability during suspend and moves some responsibility into userspace (sending feedback, finding the current buffer from the doorbell). This makes the driver simpler and less "magical".

It also fixes some sleep issues (S0ix breaking) due to improper shutdown of the device. For that a patched MEI bus driver is needed for now.

When this gets released a new iptsd release needs to be made, since the old version will not work with the new UAPI.


qzed (approved):

qzed:
Thanks!

PR #70, opened 2020-11-12
Properly fix wifi firmware crashes

jonas2515:


jonas2515:
Added some reverts for the commits which should no longer be needed now.

This needs broader testing on more devices though, especially now that we use host-sleep while suspended again.

kitakar5525:
About the ps_mode:
The ps_mode on 5 GHz AP is too unstable and almost unusable on my SB1 and S3. This is not only a fw crash issue but also a connection stability issue. So, testing by many people on various devices on both 2.4 GHz and 5 GHz AP is desirable.

About the Host Sleep:
Does S0ix still working on your SP5? Or you mean to use with bridge reset?

jonas2515:
>   Does S0ix still working on your SP5? Or you mean to use with bridge reset?

Yeah I'm always applying the bridge reset, sorry.. I'll open another PR for the bridge reset quirk, if we want to go back to host-sleep, there's no way around it.

jonas2515:
PR for the bridge-reset quirk: https://github.com/linux-surface/kernel/pull/72

drivers/net/wireless/marvell/mwifiex/sta_ioctl.c
```diff
@@ -60,9 +60,10 @@ int mwifiex_wait_queue_complete(struct mwifiex_adapter *adapter,
 	int status;
 
 	/* Wait for completion */
-	status = wait_event_interruptible_timeout(adapter->cmd_wait_q.wait,
-						  *(cmd_queued->condition),
-						  (12 * HZ));
+	status = wait_event_timeout(adapter->cmd_wait_q.wait,
```
qzed, line 63: Does that fix instabilities? If so, is there any other way we can ensure completion/timeout if a user interrupts before sending another command? In general, long uninterruptible waits should be avoided in IOCTLs/on user processes, as this makes the process uninterruptible and also unkillable. At the very least, I think `wait_event_killable_timeout` should be used.

qzed (changes requested):

kitakar5525:
Host Sleep suspend method itself won't cause fw crashes as far as I know. The problems are AP scanning after suspend and S0ix (depends on #72 to fix).
So, what about moving this change to a new PR or just changing the title of this PR to something like "mwifiex improvements for stability" ?

Also, we need to test if the method really won't cause AP scanning failure after suspend anymore. As I said in the mailing list, at a glance, it won't happen anymore.

jonas2515:
Okay, updated this one to only include the most important fix, let's do the re-enabling of host-sleep in a separate MR.

drivers/net/wireless/marvell/mwifiex/sta_ioctl.c
jonas2515: I'm not sure if this actually fixes anything, it just gets rid of a bunch of error messages on dmesg about terminated wait queues that happen sometimes. I've removed that commit for now.

jonas2515:
Okay, opened https://github.com/linux-surface/kernel/pull/89 for reenabling the host sleep, and opened https://github.com/linux-surface/kernel/pull/90 (which we maybe shouldn't merge yet) for reenabling power saving.

qzed (approved):
Thanks!

PR #71, opened 2020-11-12
mwifiex: Fixes for wifi p2p and AP mode

jonas2515:
Make wifi p2p actually work and fix a firmware crash when switching from station mode to AP mode.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/cfg80211.c
+++ b/drivers/net/wireless/marvell/mwifiex/cfg80211.c
@@ -1164,6 +1164,20 @@ mwifiex_cfg80211_change_virtual_intf(struct wiphy *wiphy,
 		return -EBUSY;
 	}
 
+	if (type == NL80211_IFTYPE_UNSPECIFIED) {
+		mwifiex_dbg(priv->adapter, INFO,
+			    "%s: no new type specified, keeping old type %d\n",
+			    dev->name, curr_iftype);
+		return 0;
+	}
+
+	if (curr_iftype == type) {
+		mwifiex_dbg(priv->adapter, INFO,
+			    "%s: interface already is of type %d\n",
+			    dev->name, curr_iftype);
+		return 0;
+	}
+
 	switch (curr_iftype) {
 	case NL80211_IFTYPE_ADHOC:
 		switch (type) {
@@ -1183,12 +1197,6 @@ mwifiex_cfg80211_change_virtual_intf(struct wiphy *wiphy,
 		case NL80211_IFTYPE_AP:
 			return mwifiex_change_vif_to_ap(dev, curr_iftype, type,
 							params);
-		case NL80211_IFTYPE_UNSPECIFIED:
-			mwifiex_dbg(priv->adapter, INFO,
-				    "%s: kept type as IBSS\n", dev->name);
-			fallthrough;
-		case NL80211_IFTYPE_ADHOC:	/* This shouldn't happen */
-			return 0;
 		default:
 			mwifiex_dbg(priv->adapter, ERROR,
 				    "%s: changing to %d not supported\n",
@@ -1214,12 +1222,6 @@ mwifiex_cfg80211_change_virtual_intf(struct wiphy *wiphy,
 		case NL80211_IFTYPE_AP:
 			return mwifiex_change_vif_to_ap(dev, curr_iftype, type,
 							params);
-		case NL80211_IFTYPE_UNSPECIFIED:
-			mwifiex_dbg(priv->adapter, INFO,
-				    "%s: kept type as STA\n", dev->name);
-			fallthrough;
-		case NL80211_IFTYPE_STATION:	/* This shouldn't happen */
-			return 0;
 		default:
 			mwifiex_dbg(priv->adapter, ERROR,
 				    "%s: changing to %d not supported\n",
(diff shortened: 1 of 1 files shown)


drivers/net/wireless/marvell/mwifiex/cfg80211.c
```diff
@@ -962,6 +962,138 @@ mwifiex_init_new_priv_params(struct mwifiex_private *priv,
...
+		case NL80211_IFTYPE_P2P_GO:
+			adapter->curr_iface_comb.p2p_intf++;
+			break;
+		default:
+			/* This should be dead code; checked above */
+			mwifiex_dbg(adapter, ERROR, "type not supported\n");
+			break;
+		}
+	}
+}
```
qzed, line 1095: I think it would be cleaner to split this into two functions, one for incrementing and one for decrementing (removing the need for the special `-1` value). Also you can split out a third function that updates the counters based on an input value, i.e. `update(type, value)` and then use `update(type, +1)` and `update(type, -1)` inside some small wrappers for inc and dec (basically just for naming).

drivers/net/wireless/marvell/mwifiex/cfg80211.c
```diff
@@ -962,6 +962,138 @@ mwifiex_init_new_priv_params(struct mwifiex_private *priv,
...
+{
+	switch (old_iftype) {
+	case NL80211_IFTYPE_ADHOC:
+		switch (new_iftype) {
+		case NL80211_IFTYPE_STATION:
+			return true;
+		case NL80211_IFTYPE_P2P_CLIENT:
+		case NL80211_IFTYPE_P2P_GO:
+			return adapter->curr_iface_comb.p2p_intf ==
+			       adapter->iface_limit.p2p_intf;
```
qzed, line 978: Are these checks correct? In the original code you have
```c
if (adapter->curr_iface_comb.p2p_intf ==
    adapter->iface_limit.p2p_intf) {
	mwifiex_dbg(adapter, ERROR,
		    "cannot create multiple P2P ifaces\n");
	return -1;
}
```
so shouldn't this be
```c
return adapter->curr_iface_comb.p2p_intf !=
       adapter->iface_limit.p2p_intf;
```
instead?

qzed (changes requested):
Can't comment on the functionality, but code looks ok. I'm not quite sure what to think of splitting all the transition checks out of the individual transition functions (I generally like to keep those things as local as possible), but I guess in this case it does simplify the transitioning code quite a bit, so I'm okay with that.

drivers/net/wireless/marvell/mwifiex/cfg80211.c
jonas2515: Ouch, nice catch, indeed! That's what reviews are for, aren't they ;)

jonas2515:
Alright, I've updated this to target 5.11 and added a few more fixes. IMHO the quality of this code is quite embarrassing for Marvell, it looks almost like the whole wifi p2p functionality was never tested.

drivers/net/wireless/marvell/mwifiex/cfg80211.c
jonas2515: Went with only one helper function that gets passed `+1`/`-1` directly now, adding increase/decrease wrappers seemed a little too much IMHO.

drivers/net/wireless/marvell/mwifiex/cfg80211.c
jonas2515: Fixed that now

drivers/net/wireless/marvell/mwifiex/cfg80211.c
```diff
@@ -1214,88 +1288,65 @@ mwifiex_cfg80211_change_virtual_intf(struct wiphy *wiphy,
...
-		default:
-			mwifiex_dbg(priv->adapter, ERROR,
-				    "%s: changing to %d not supported\n",
-				    dev->name, type);
+			return mwifiex_change_vif_to_p2p(dev, curr_iftype,
+							 type, params);
+		case NL80211_IFTYPE_AP:
+			return mwifiex_change_vif_to_ap(dev, curr_iftype, type,
+							params);
+		default: /* Should never get here, handled by is_vif_type_change_allowed() */
```
qzed, line 1344: You could add a `WARN(true, "unchecked interface transition: %d to %d", curr_iftype, type);` here. That way if this ever happens it gets noticed ant we know what change caused the warning.

qzed (approved):
One small debugability thing again, but otherwise good as usual!

drivers/net/wireless/marvell/mwifiex/cfg80211.c
jonas2515: Good idea, went with `mwifiex_dbg` for consistency and used a goto to avoid duplicating it everywhere.

qzed: I'm not entirely sure if `mwifiex_dbg` is the best function for that since it seems to use `dev_info`/`pr_info` under the hood, maybe something stronger that sticks out in dmesg would be better. On the other hand though, they essentially seem to have replicated `dev_{info,warn,err,...}` with that so I don't know what to think of that macro...

Probably better to stick with it for consistency.

PR #72, opened 2020-11-12
mwifiex: Add quirk resetting the PCI bridge on MS Surface devices

jonas2515:
The most recent firmware of the 88W8897 card reports a hardcoded LTR
value to the system during initialization, probably as an (unsuccessful)
attempt of the developers to fix firmware crashes. This LTR value
prevents most of the Microsoft Surface devices from entering deep
powersaving states (either platform C-State 10 or S0ix state), because
the exit latency of that state would be higher than what the card can
tolerate.

Turns out the card works just the same (including the firmware crashes)
no matter if that hardcoded LTR value is reported or not, so it's kind
of useless and only prevents us from saving power.

To get rid of those hardcoded LTR reports, it's possible to reset the
PCI bridge device after initializing the cards firmware. I'm not exactly
sure why that works, maybe the power management subsystem of the PCH
resets its stored LTR values when doing a function level reset of the
bridge device. Doing the reset once after starting the wifi firmware
works very well, probably because the firmware only reports that LTR
value a single time during firmware startup.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
+++ b/drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
@@ -12,6 +12,7 @@
  */
 #define QUIRK_FW_RST_WSID_S3	BIT(1)
 #define QUIRK_NO_BRIDGE_D3	BIT(2)
+#define QUIRK_DO_FLR_ON_BRIDGE	BIT(3)
 
 void mwifiex_initialize_quirks(struct pcie_service_card *card);
 int mwifiex_pcie_reset_d3cold_quirk(struct pci_dev *pdev);
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -1753,8 +1753,20 @@ mwifiex_pcie_send_boot_cmd(struct mwifiex_adapter *adapter, struct sk_buff *skb)
 static int mwifiex_pcie_init_fw_port(struct mwifiex_adapter *adapter)
 {
 	struct pcie_service_card *card = adapter->card;
+	struct pci_dev *pdev = card->dev;
+	struct pci_dev *parent_pdev = pci_upstream_bridge(pdev);
 	const struct mwifiex_pcie_card_reg *reg = card->pcie.reg;
 	int tx_wrap = card->txbd_wrptr & reg->tx_wrap_mask;
+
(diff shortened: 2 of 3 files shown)


drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
```diff
@@ -12,6 +12,7 @@
  */
 #define QUIRK_FW_RST_WSID_S3	BIT(1)
 #define QUIRK_NO_BRIDGE_D3	BIT(2)
+#define QUIRK_DO_FLR_ON_BRIDGE	BIT(4)
```
qzed, line 15: Any reason for skipping `BIT(3)`?

qzed (changes requested):

kitakar5525:
Thanks to your work on fixing fw crashes, hopefully this bridge reset won't cause fw crashes anymore on all the devices.

So, what about applying this quirk to all the Surface devices? (No effect on Surface 3 and I guess SP3 is the same as well, though. So, you can skip applying for S3 and SP3 for now I think.)

Also, if you intend to revert suspend method back to Host Sleep, this quirk is required for S0ix on suspend.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
jonas2515: Whoops, I had another quirk for disabling ASPM L1.2 before that, which is obviously no longer needed now.

jonas2515:
> So, what about applying this quirk to all the Surface devices? (No effect on Surface 3 and I guess SP3 is the same as well, though. So, you can skip applying for S3 and SP3 for now I think.)
>
> Also, if you intend to revert suspend method back to Host Sleep, this quirk is required for S0ix on suspend.

Sounds good, indeed I didn't think of the consequences for other devices when re-enabling Host Sleep on all devices.

drivers/net/wireless/marvell/mwifiex/pcie_quirks.h
jonas2515: Changed to `BIT(3)`

jonas2515:
This targets branch `v5.11-surface-devel` now.

I've also updated this to add the quirk to all Surface devices except Surface 3 and Pro 3, as suggested by @kitakar5525.

qzed (approved):
Thanks, looks good!

PR #73, opened 2020-12-10
Initial camera support for v5.10

qzed:
CC @djrscally, I believe you should have access to this repo now, so feel free to push if anything is missing. Also feel free to add/adapt the last three commits to your patchset.

I have yet to test this fully. Will merge after that and your approval.


qzed:
Ah, the sensor drivers are still missing here. I'll add them in a sec.

qzed:
Tested and confirmed working on my SB2. Would be neat if we can get a couple of other devices tested and supported, we likely have to add some IDs for that to the int3472 driver.

djrscally:
Sorry, forgot to comment here.

I've tested this on my Miix-510 and Go2 and it works fine. Only annoyance is on the Go2, the OV8865 uses a physical PMIC which isn't supported yet (soon!) so you have to remove INT347A from supported_sensors for that device or **neither** camera will work - this is due to how the ipu3-cio2 driver decides when to finalise setup...it won't make the cameras available as v4l2 subdevs until ALL of the sensors you tell it are present have probed.

Everything else was fine - I'll make time tonight to update the Wiki with that info.

qzed:
Thanks, I'll remove `INT347A` for the time being then.

I'm not merging this immediately as I want to have a look at the DSDTs for other devices before, I think we can also enable support for the OV5693 on the Surface Pros and Laptops with that sensor if we find the right IDs.

PR #74, opened 2020-12-22
Add MSHW0184 detection and proximity sensor support to the apds9960 drivers

MaxLeiter:
Adds support for Surface Book 3, Surface Laptop 3, and the Surface Pro 7.

The ACPI table additions are currently in the testing stage of the iio subtree, ref: http://lkml.iu.edu/hypermail/linux/kernel/2012.2/03697.html

Closes https://github.com/linux-surface/linux-surface/issues/121

The proximity-near-level code documentation can be found in the `drivers/light/iio/vcnl4000.c` driver implementation and at https://www.kernel.org/doc/Documentation/ABI/testing/sysfs-bus-iio

### DIFF
--- a/drivers/iio/light/apds9960.c
+++ b/drivers/iio/light/apds9960.c
@@ -8,6 +8,7 @@
  * TODO: gesture + proximity calib offsets
  */
 
+#include <linux/acpi.h>
 #include <linux/module.h>
 #include <linux/init.h>
 #include <linux/interrupt.h>
@@ -145,6 +146,9 @@ struct apds9960_data {
 
 	/* gesture buffer */
 	u8 buffer[4]; /* 4 8-bit channels */
+
+	/* proximity near level */
+	uint32_t near_level;
 };
 
 static const struct reg_default apds9960_reg_defaults[] = {
@@ -273,6 +277,15 @@ static const struct iio_event_spec apds9960_als_event_spec[] = {
 
 static const unsigned long apds9960_scan_masks[] = {0xf, 0};
 
+static const struct iio_chan_spec_ext_info adps9960_ext_info[] = {
+	{
+		.name = "near_level",
+		.shared = IIO_SEPARATE,
+		.read = adps9960_read_near_level,
+	},
+	{}
+};
+
 static const struct iio_chan_spec apds9960_channels[] = {
 	{
 		.type = IIO_PROXIMITY,
@@ -301,6 +314,7 @@ static const struct iio_chan_spec apds9960_channels[] = {
 		.address = APDS9960_REG_ALS_CHANNEL(CLEAR),
 		.modified = 1,
 		.scan_index = -1,
+		.ext_info = adps9960_ext_info,
 
 		.event_spec = apds9960_als_event_spec,
 		.num_event_specs = ARRAY_SIZE(apds9960_als_event_spec),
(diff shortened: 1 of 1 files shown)


drivers/iio/light/apds9960.c
```diff
@@ -145,6 +146,9 @@ struct apds9960_data {
 
 	/* gesture buffer */
 	u8 buffer[4]; /* 4 8-bit channels */
+
+	/* proximity near level */
+	uint32_t near_level;
```
qzed, line 151: This should be `u32` instead of `uint32_t`.

drivers/iio/light/apds9960.c
```diff
@@ -273,6 +277,15 @@ static const struct iio_event_spec apds9960_als_event_spec[] = {
 
 static const unsigned long apds9960_scan_masks[] = {0xf, 0};
 
+static const struct iio_chan_spec_ext_info adps9960_ext_info[] = {
+	{
+		.name = "near_level",
```
qzed, line 282: I assume this should be `nearlevel` and not `near_level`? At least according to `Documentation/ABI/testing/sysfs-bus-iio-proximity` and the only other driver that provides such a thing.

drivers/iio/light/apds9960.c
```diff
@@ -301,6 +314,7 @@ static const struct iio_chan_spec apds9960_channels[] = {
 		.address = APDS9960_REG_ALS_CHANNEL(CLEAR),
 		.modified = 1,
 		.scan_index = -1,
+		.ext_info = adps9960_ext_info,
```
qzed, line 317: Those should be two different commits: One for adding the `near_level` attribute and one for adding the `MSHW0184` match (which you've already submitted upstream).

drivers/iio/light/apds9960.c
```diff
@@ -356,6 +370,13 @@ static const struct reg_field apds9960_reg_field_enable_ges =
 static const struct reg_field apds9960_reg_field_enable_pxs =
 				REG_FIELD(APDS9960_REG_ENABLE, 2, 2);
 
+static ssize_t adps9960_read_near_level(struct iio_dev *indio_dev, uintptr_t priv, const struct iio_chan_spec *chan, char *buf)
+{
+	struct apds9960_data *data = iio_priv(indio_dev);
+
+	return sprintf(buf, "%u\n", data->near_level);
```
qzed, line 377: This should be `sysfs_emit` instead of `sprintf`.

drivers/iio/light/apds9960.c
```diff
@@ -1018,6 +1039,10 @@ static int apds9960_probe(struct i2c_client *client,
 		return PTR_ERR(data->regmap);
 	}
 
+	if (device_property_read_u32(&client->dev, "proximity-near-level", &data->near_level) < 0) {
+		data->near_level = 0;
+	}
```
qzed, line 1044: Where do you get this attribute from? I can't see that anywhere in the DSDT/acpidump. The only reference to it that I can find is in some device-tree files for another sensor. At this point, I have to ask: Do we even need that? If so, for what?

As far as I can tell, you currently a) don't have any way to get a sane value for this attribute (so on the SB3 it will always be zero) and b) probably don't have any user-space tool/library that depends on this, as you should have noticed the typo above if you had. So based on that, it doesn't make any sense to me why we should currently add this.

Also, in Linux coding style it is preferred to omit braces here and the line exceeds 100 columns. FYI, the `checkpatch.pl` script of the kernel can spot these things for you, especially if you run it with the `--strict` option.

qzed (changes requested):
I've added some comments. In addition, there should be spaces in the commit prefix, i.e. `iio: light: apds9960: ...`

archseer:
Quick note: we will need to add the required config changes to `surface.config` , I think most distros don't enable this sensor by default (it's disabled on fedora)

qzed:
Right, thanks for reminding me!

qzed:
I've applied the ACPI ID part as 23b180c407ed24eca2c2188401e264be9b123215.

qzed:
I'm going to close this as this is targeting an outdated branch and there haven't been any updates lately. Also the main part is now upstream (and has already been integrated previously). Feel free to submit a new PR if you think that there's anything still missing.

PR #75, opened 2020-12-28
ov5693: Add user controls to flip image

fabwu:
Add V4L2_CID_HFLIP and V4L2_CID_VFLIP to support flipping the image on the sensor side.

I added some simple test cases to the ipu3 pipeline in libcamera and the image does change direction (tested on SB2). A full implementation for the ipu3 pipeline will follow later.

### DIFF
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -828,6 +828,29 @@ static int ov5693_t_focus_rel(struct v4l2_subdev *sd, s32 value)
 	return ov5693_t_focus_abs(sd, dev->focus + value);
 }
 
+static int ov5693_update_hflip(struct v4l2_subdev *sd)
+{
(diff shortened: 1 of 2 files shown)


qzed (changes requested):
Thanks, looks good overall, but there are a couple of style issues (e.g. spaces instead of tabs for indentation and things like that). Would you mind running `checkpatch --strict` on the patch and fix those?

fabwu:
@qzed Thanks for the review. I fixed the style issues in my changes but the driver has more of them. Should I create a separate PR to fix these issues as well?

qzed (approved):

qzed:
Thanks!

> I fixed the style issues in my changes but the driver has more of them. Should I create a separate PR to fix these issues as well?

That'd be great if you have the time for it.

PR #77, opened 2021-01-11
ov5693: Expose sensor orientation and rotation

fabwu:
Expose the V4L2_CID_CAMERA_ORIENTATION and V4L2_CID_CAMERA_SENSOR_ROTATION controls to libcamera.

Currently, the values are hard-coded but we could extract them from the SSDB buffer.

### DIFF
--- a/drivers/media/pci/intel/ipu3/cio2-bridge.c
+++ b/drivers/media/pci/intel/ipu3/cio2-bridge.c
@@ -29,6 +29,7 @@ static const struct cio2_sensor_config cio2_supported_sensors[] = {
 static const struct cio2_property_names prop_names = {
 	.clock_frequency = "clock-frequency",
 	.rotation = "rotation",
+	.orientation = "orientation",
 	.bus_type = "bus-type",
 	.data_lanes = "data-lanes",
 	.remote_endpoint = "remote-endpoint",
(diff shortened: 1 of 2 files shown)


fabwu:
@qzed It might be better if you give me access to the ipu3 branch so I don't have to bother you with these experimental PR.

djrscally:
Not a comment on this change, but wrt reading from SSDB I think it should be read from the software_nodes created by cio2-bridge instead, as in `v4l2_fwnode_device_parse()`:

```c
u32 val;

...

	ret = fwnode_property_read_u32(fwnode, "rotation", &val);
	if (!ret) {
		if (val >= 360) {
			dev_warn(dev, "Unsupported device rotation: %u\n", val);
			return -EINVAL;
		}

		props->rotation = val;
		dev_dbg(dev, "device rotation: %u\n", val);
	}
```

Particularly as it seems like those SSDB values might not be reliable, which means we'll end up having platform specific settings somewhere most likely.

qzed:
@fabwu You should have access now.

fabwu:
> Particularly as it seems like those SSDB values might not be reliable, which means we'll end up having platform specific settings somewhere most likely.

Yes, that's annoying. I'll add your example and hope the libcamera rotation patch gets merged soon. Then we can run some tests with actual images and see how reliable these values are.

qzed:
Looks good to me so far. Feel free to merge it (I think you should have the rights to do that, let me know if not).

@djrscally So the ideal way would be the cio2-bridge reading SSDB (performing any necessary fix-ups on that if it's not reliably) and creating the attributes?

djrscally:
> Looks good to me so far. Feel free to merge it (I think you should have the rights to do that, let me know if not).
> 
> @djrscally So the ideal way would be the cio2-bridge reading SSDB (performing any necessary fix-ups on that if it's not reliably) and creating the attributes?



> Looks good to me so far. Feel free to merge it (I think you should have the rights to do that, let me know if not).
> 
> @djrscally So the ideal way would be the cio2-bridge reading SSDB (performing any necessary fix-ups on that if it's not reliably) and creating the attributes?

Yes; it's already reading the SSDB values and creating the property against the firmware node; but @fabwu pointed out his sensor's upside down with SSDB reporting rotation as 0...so that ain't right. Most sensors I've seen are the right way round and reporting 0.

qzed:
Thanks for the explanation! Bit annoying that we can't rely on those, so makes sense to keep that in one place and use fwnodes.

PR #79, opened 2021-01-22
ipu3: Image rotation support

fabwu:
This PR adds rotation/orientation support for cio2-bridge and ov5693.

The cio2-bridge parses the sensor orientation and rotation from the ACPI buffers and expose them as fwnodes. The ov5693 reads the values from the fwnodes and create the controls. Then libcamera picks up the controls and rotate the image in the pipeline. 

To test the whole thing you need this PR and you have to apply this [patch](https://patchwork.libcamera.org/patch/10957/) to libcamera.

Does anyone have a better idea for [this part](https://github.com/fabwu/linux-surface-kernel/commit/3c644c43e4860cacec1d8523728f3373a9fbc5fb#diff-99ed78a9282d84363094cde07371167771226cd506695acab79c3f42044db146R76-R94)? Converting the values like that feels a bit ugly...

### DIFF
--- a/drivers/media/pci/intel/ipu3/cio2-bridge.c
+++ b/drivers/media/pci/intel/ipu3/cio2-bridge.c
@@ -73,21 +73,36 @@ out_free_buff:
 	return ret;
 }
 
-static u32 cio2_bridge_parse_rotation(u8 rotation)
+static u32 cio2_bridge_parse_rotation(struct cio2_sensor *sensor)
 {
-	if (rotation == 1)
+	switch (sensor->ssdb.degree) {
+	case CIO2_SENSOR_ROTATION_NORMAL:
+		return 0;
+	case CIO2_SENSOR_ROTATION_INVERTED:
 		return 180;
-	return 0;
+	default:
+		dev_warn(&sensor->adev->dev,
+			 "Unknown rotation %d. Assume 0 degree rotation\n",
+			 sensor->ssdb.degree);
+		return 0;
+	}
 }
 
-static enum v4l2_fwnode_orientation cio2_bridge_parse_orientation(u8 panel)
+static enum v4l2_fwnode_orientation cio2_bridge_parse_orientation(struct cio2_sensor *sensor)
 {
-	switch (panel) {
-	case 4:
+	switch (sensor->pld->panel) {
+	case CIO2_PLD_PANEL_FRONT:
 		return V4L2_FWNODE_ORIENTATION_FRONT;
-	case 5:
+	case CIO2_PLD_PANEL_BACK:
 		return V4L2_FWNODE_ORIENTATION_BACK;
+	case CIO2_PLD_PANEL_TOP:
+	case CIO2_PLD_PANEL_LEFT:
+	case CIO2_PLD_PANEL_RIGHT:
+	case CIO2_PLD_PANEL_UNKNOWN:
+		return V4L2_FWNODE_ORIENTATION_EXTERNAL;
 	default:
+		dev_warn(&sensor->adev->dev, "Unknown _PLD panel value %d\n",
+			 sensor->pld->panel);
 		return V4L2_FWNODE_ORIENTATION_EXTERNAL;
 	}
 }
@@ -100,8 +115,8 @@ static void cio2_bridge_create_fwnode_properties(
 	u32 rotation;
 	enum v4l2_fwnode_orientation orientation;
 
-	rotation = cio2_bridge_parse_rotation(sensor->ssdb.degree);
-	orientation = cio2_bridge_parse_orientation(sensor->pld->panel);
+	rotation = cio2_bridge_parse_rotation(sensor);
+	orientation = cio2_bridge_parse_orientation(sensor);
 
 	sensor->prop_names = prop_names;
 
--- a/drivers/media/pci/intel/ipu3/cio2-bridge.c
+++ b/drivers/media/pci/intel/ipu3/cio2-bridge.c
@@ -29,6 +29,7 @@ static const struct cio2_sensor_config cio2_supported_sensors[] = {
 static const struct cio2_property_names prop_names = {
 	.clock_frequency = "clock-frequency",
 	.rotation = "rotation",
+	.orientation = "orientation",
 	.bus_type = "bus-type",
 	.data_lanes = "data-lanes",
 	.remote_endpoint = "remote-endpoint",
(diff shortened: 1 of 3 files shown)


drivers/media/pci/intel/ipu3/cio2-bridge.c
```diff
@@ -72,11 +73,36 @@ static int cio2_bridge_read_acpi_buffer(struct acpi_device *adev, char *id,
...
+	switch (panel) {
+	case 4:
+		return V4L2_FWNODE_ORIENTATION_FRONT;
+	case 5:
+		return V4L2_FWNODE_ORIENTATION_BACK;
+	default:
+		return V4L2_FWNODE_ORIENTATION_EXTERNAL;
+	}
+}
+
```
qzed, line 94: We don't have any spec for this, right? Maybe print a warning or debug for any unknown value?

qzed (approved):
Looks good to me! Maybe add a print statement for any unknown or unexpected values. I can't comment much on the functionality itself, so I leave that to @djrscally.

drivers/media/pci/intel/ipu3/cio2-bridge.c
fabwu: The rotation is reverse engineered so 1 == 180 is just a guess but a comment is indeed necessary.

For the orientation we have [this spec](https://uefi.org/sites/default/files/resources/ACPI_6_3_final_Jan30.pdf) on page 351 so I'll add a comment for that as well.

What do you think about the switch-statement? Is there a better way of doing the conversion?

djrscally: I think you'd probably find / need to add macros for the 4/5 values for the _PLD panel field when this goes upstream. The switch itself is fine imo.

I'm surprised to find 1==180, as it sort of precludes a sensor being mounted at 90 degrees or something...I was expecting something like  1 = 90 degrees, 2 = 180. Anyway, I think that probably ought to return the input value in the case that it's not 1, otherwise if we do find some other value in there it'll be masked here.

qzed: > For the orientation we have [this spec](https://uefi.org/sites/default/files/resources/ACPI_6_3_final_Jan30.pdf) on page 351 so I'll add a comment for that as well.

Thanks for the link! Maybe mention that this value comes from the ACPI PLD buffer, then people can find the spec for that themselves. Also since anything except front and back seem to be directly translatable from ACPI, I think you should go with a warning here (as opposed to a debug print).

> I think you'd probably find / need to add macros for the 4/5 values for the _PLD panel field when this goes upstream. The switch itself is fine imo.

Yeah, I think those should be constants if they're defined in ACPI. Maybe also for the rotation.

> What do you think about the switch-statement? Is there a better way of doing the conversion?

I agree with @djrscally. I think switch is probably the best option here. A lookup-table would be an alternative, but doesn't make much sense here, I think.

> I'm surprised to find 1==180, as it sort of precludes a sensor being mounted at 90 degrees or something...I was expecting something like 1 = 90 degrees, 2 = 180.

I assume we don't know any device that has a 90 degree sensor so we could check this?

> Anyway, I think that probably ought to return the input value in the case that it's not 1, otherwise if we do find some other value in there it'll be masked here.

I'm not entirely sure if returning the input value is a good thing, but there should be something to help us figure out if there's an unknown value (that's why I proposed warning or debug messages).

djrscally: > Maybe also for the rotation.

Yeah, they could just got in cio2-bridge.h

> I assume we don't know any device that has a 90 degree sensor so we could check this?

Not to my knowledge, I don't recall anything except a 180 rotation, I was just going off what Windows apparently expects from the rotation field in _PLD: https://docs.microsoft.com/en-us/windows-hardware/drivers/stream/camera-device-orientation.

Then again, given the field is called "degree" you'd really expect it to be in degrees if anything!

> I'm not entirely sure if returning the input value is a good thing, but there should be something to help us figure out if there's an unknown value (that's why I proposed warning or debug messages).

Fair enough; a debug print is good for me too then

djrscally (approved):
This looks good, I agree with adding a debug print for unexpected values, and maybe make the constants into macros when it goes upstream. Thanks @fabwu

drivers/media/pci/intel/ipu3/cio2-bridge.c
qzed: > Then again, given the field is called "degree" you'd really expect it to be in degrees if anything!

Yeah... let's just hope the 0/1 thing is consistent for now and we don't end up with other devices actually having a rotation in degrees there. Might be a bit messy handling that.

fabwu: It's also strange that `degree` is only 8-bit so it cannot encode e.g. 270 degree directly.

I'll try to go through the SSDB buffers and find all values for degree then we have a better picture.

fabwu: I grepped the ACPI info from [here](https://github.com/kitakar5525/surface-ipu3-cameras/tree/master/misc/dump_intel_ipu_data/results) and most of the sensors have 0 or 1 in the degree field (SSDB) and 0 in the roation field (PLD).

There are some quirks though:
- Switch_SA5-271 has the degree field set to 180 (yey!)
- Surface_Book_3 and Surface_Pro_7 report 2 (i.e. 90 degree) in the PLD roation field for the IR-Sensor

I think it's save to treat degree as a "binary" variable for now and treat the three devices separately.

qzed: > There are some quirks though:
> 
> * Switch_SA5-271 has the degree field set to 180 (yey!)

Whelp, of course there are... would be too easy otherwise, would it. I think we can assume though that it's either 0, 1, 2, (maybe 3?) and 90, 180. So still possible to somewhat sanely handle those.

>  * Surface_Book_3 and Surface_Pro_7 report 2 (i.e. 90 degree) in the PLD roation field for the IR-Sensor

Hmm, too bad those are IPU4 devices. But at least we now know that we can maybe check these values in the future.

qzed (approved):

drivers/media/pci/intel/ipu3/cio2-bridge.c
djrscally: > There are some quirks though:
>
>    Switch_SA5-271 has the degree field set to 180 (yey!)

Oh good grief! 

> I think it's save to treat degree as a "binary" variable for now and treat the three devices separately.

Yeah, sounds sensible to me.

djrscally (approved):

qzed:
@djrscally, @fabwu, feel free to merge this, I think you should have the proper access rights (of not, let me know).

PR #81, opened 2021-02-08
Sync with IPTS module repo

StollD:
These changes have been sitting in the module repo for a while now, and nobody has screamed at me yet, so lets sync them into the patches.

This has a bit of everything:
 - You can now reset the sensor from userspace, the master branch of iptsd already has support for this
 - It doesnt require patches to the MEI bus anymore, so at least on gen4 - gen6 IPTS will work as a DKMS module. Gen7 will need upstreaming of the device ID's
 - A lot of formatting stuff for eventual upstreaming (not that I want to work on that now, but it doesnt hurt).

### DIFF
--- a/drivers/misc/ipts/context.h
+++ b/drivers/misc/ipts/context.h
@@ -45,4 +45,3 @@ struct ipts_context {
 };
 
 #endif /* _IPTS_CONTEXT_H_ */
-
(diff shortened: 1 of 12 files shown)


qzed (approved):
Looks good from my point of view. Feel free to merge this!

PR #82, opened 2021-02-13
Updates and Improvements to ov5693

djrscally:
Hello

Here's a long batch of updates to the ov5693 driver. It's mostly tidying, with some functional changes:

- pm_runtime support replacing the existing power functions
- Fixed control setting (which previously was broken)
- Added vblank control (needed in latest versions of libcamera)

I'm planning on working on this quite a lot (meaning there'll be more to come), but I think @fabwu at least was too, so wanted to update you with what I'd done so far so we're not duplicating effort, and also because without the vblank control the driver won't work for anyone using latest versions of libcamera (already the cause of comments on #91)

Note that with later versions of libcamera this will now seem broken, when you stream in qcam it'll insta-max exposure which means the image will be white and FPS about 1-2 maximum. That's due to libcamera maxing them (which previously just wasn't being applied because nothing in the driver was doing it). You can use `sudo v4l2-ctl -d /dev/v4l-subdevn --set-ctrl exposure=123` to drop it to a reasonable value (which'll fix the FPS too). Though note that without a [recent patch](https://lists.libcamera.org/pipermail/libcamera-devel/2021-February/017258.html), it'll re-max it 2 frames later, so it's a bit annoying.

### DIFF
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -71,9 +71,9 @@ MODULE_PARM_DESC(up_delay,
 #define OV5693_EXPOSURE_GAIN_MANUAL_REG		0x3509
 
 #define OV5693_GAIN_CTRL_H_REG			0x3504
-#define OV5693_GAIN_CTRL_H(v)			(((v) & GENMASK(9, 8)) >> 8)
+#define OV5693_GAIN_CTRL_H(v)			((v >> 4) & GENMASK(2, 0))
 #define OV5693_GAIN_CTRL_L_REG			0x3505
-#define OV5693_GAIN_CTRL_L(v)			((v) & GENMASK(7, 0))
+#define OV5693_GAIN_CTRL_L(v)			((v << 4) & GENMASK(7, 4))
 
 #define OV5693_FORMAT1_REG			0x3820
 #define OV5693_FORMAT1_FLIP_VERT_ISP_EN		BIT(2)
@@ -889,9 +889,13 @@ static int ov5693_analog_gain_configure(struct ov5693_device *sensor, u32 gain)
 {
 	int ret;
 
-	/* Analog gain */
+	/*
+	 * As with exposure, the lowest 4 bits are fractional bits. Setting
+	 * those is not supported, so we have a tiny bit of bit shifting to
+	 * do.
+	 */
 	ret = ov5693_write_reg(sensor->i2c_client, OV5693_8BIT,
-				OV5693_AGC_L, gain & 0xff);
+				OV5693_AGC_L, OV5693_GAIN_CTRL_L(gain));
 	if (ret) {
 		dev_err(&sensor->i2c_client->dev, "%s: write %x error, aborted\n",
 			__func__, OV5693_AGC_L);
@@ -899,7 +903,7 @@ static int ov5693_analog_gain_configure(struct ov5693_device *sensor, u32 gain)
 	}
 
 	ret = ov5693_write_reg(sensor->i2c_client, OV5693_8BIT,
-				OV5693_AGC_H, (gain >> 8) & 0xff);
+				OV5693_AGC_H, OV5693_GAIN_CTRL_H(gain));
 	if (ret) {
 		dev_err(&sensor->i2c_client->dev, "%s: write %x error, aborted\n",
 			__func__, OV5693_AGC_H);
@@ -1674,10 +1678,10 @@ static int ov5693_init_controls(struct ov5693_device *ov5693)
 
 	ov5693->ctrls.analogue_gain = v4l2_ctrl_new_std(&ov5693->ctrl_handler,
 							ops, V4L2_CID_ANALOGUE_GAIN,
-							1, 1023, 1, 128);
+							1, 127, 1, 8);
 	ov5693->ctrls.digital_gain = v4l2_ctrl_new_std(&ov5693->ctrl_handler, ops,
 						       V4L2_CID_DIGITAL_GAIN, 1,
-						       3999, 1, 1000);
+						       4095, 1, 1024);
 
 	/* Flip */
 
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -63,11 +63,11 @@ MODULE_PARM_DESC(up_delay,
 /* Exposure/gain */
 
 #define OV5693_EXPOSURE_CTRL_HH_REG		0x3500
-#define OV5693_EXPOSURE_CTRL_HH(v)		(((v) & GENMASK(18, 16)) >> 16)
+#define OV5693_EXPOSURE_CTRL_HH(v)		(((v) & GENMASK(14, 12)) >> 12)
 #define OV5693_EXPOSURE_CTRL_H_REG		0x3501
-#define OV5693_EXPOSURE_CTRL_H(v)		(((v) & GENMASK(15, 8)) >> 8)
+#define OV5693_EXPOSURE_CTRL_H(v)		(((v) & GENMASK(11, 4)) >> 4)
 #define OV5693_EXPOSURE_CTRL_L_REG		0x3502
-#define OV5693_EXPOSURE_CTRL_L(v)		((v) & GENMASK(7, 0))
+#define OV5693_EXPOSURE_CTRL_L(v)		(((v) & GENMASK(3, 0)) << 4)
 #define OV5693_EXPOSURE_GAIN_MANUAL_REG		0x3509
 
 #define OV5693_GAIN_CTRL_H_REG			0x3504
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -1492,6 +1492,12 @@ static int ov5693_s_stream(struct v4l2_subdev *sd, int enable)
 		}
 	}
 
+	ret = __v4l2_ctrl_handler_setup(&dev->ctrl_handler);
+	if (ret) {
+		power_down(sd);
+		return ret;
+	}
+
 	ret = ov5693_write_reg(client, OV5693_8BIT, OV5693_SW_STREAM,
 			       enable ? OV5693_START_STREAMING :
 			       OV5693_STOP_STREAMING);
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -770,30 +770,35 @@ static int ov5693_t_focus_rel(struct v4l2_subdev *sd, s32 value)
 
 static int ov5693_get_exposure(struct ov5693_device *sensor)
 {
-	u16 reg_v, reg_v2;
+	u32 exposure = 0;
+	u16 tmp;
 	int ret = 0;
 
 	/* get exposure */
 	ret = ov5693_read_reg(sensor->i2c_client, OV5693_8BIT,
 			      OV5693_EXPOSURE_L,
-			      &reg_v);
+			      &tmp);
 	if (ret)
 		return ret;
+
+	exposure |= ((tmp >> 4) & 0b1111);
 
 	ret = ov5693_read_reg(sensor->i2c_client, OV5693_8BIT,
 			      OV5693_EXPOSURE_M,
-			      &reg_v2);
+			      &tmp);
 	if (ret)
 		return ret;
 
-	reg_v += reg_v2 << 8;
+	exposure |= (tmp << 4);
 	ret = ov5693_read_reg(sensor->i2c_client, OV5693_8BIT,
 			      OV5693_EXPOSURE_H,
-			      &reg_v2);
+			      &tmp);
 	if (ret)
 		return ret;
 
-	printk("exposure set to: %u\n", reg_v + (((u32)reg_v2 << 16)));
+	exposure |= (tmp << 12);
+
+	printk("exposure set to: %u\n", exposure);
 	return ret;
 }
 
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -1266,12 +1266,8 @@ static int ov5693_remove(struct i2c_client *client)
 {
 	struct v4l2_subdev *sd = i2c_get_clientdata(client);
 	struct ov5693_device *ov5693 = to_ov5693_sensor(sd);
-	unsigned int i = OV5693_NUM_SUPPLIES;
 
 	dev_info(&client->dev, "%s...\n", __func__);
-
-	while (i--)
-		regulator_put(ov5693->supplies[i].consumer);
 
 	v4l2_async_unregister_subdev(sd);
 
@@ -1401,7 +1397,7 @@ static int ov5693_get_regulators(struct ov5693_device *ov5693)
 	for (i = 0; i < OV5693_NUM_SUPPLIES; i++)
 		ov5693->supplies[i].supply = ov5693_supply_names[i];
 
-	return regulator_bulk_get(&ov5693->client->dev,
+	return devm_regulator_bulk_get(&ov5693->client->dev,
 				       OV5693_NUM_SUPPLIES,
 				       ov5693->supplies);
 }
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -767,6 +767,42 @@ static int ov5693_sw_standby(struct ov5693_device *ov5693, bool standby)
 			       standby ? OV5693_STOP_STREAMING : OV5693_START_STREAMING);
 }
 
+static int ov5693_sw_reset(struct ov5693_device *ov5693)
+{
+	return ov5693_write_reg(ov5693->client, OV5693_8BIT, OV5693_SW_RESET,
+				0x01);
+}
+
+static int ov5693_sensor_init(struct ov5693_device *ov5693)
+{
+	struct i2c_client *client = ov5693->client;
+	int ret = 0;
+
+	ret = ov5693_sw_reset(ov5693);
+	if (ret) {
+		dev_err(&client->dev, "ov5693 reset err.\n");
+		return ret;
+	}
+
+	ret = ov5693_write_reg_array(client, ov5693_global_setting);
+	if (ret) {
+		dev_err(&client->dev, "ov5693 write register err.\n");
+		return ret;
+	}
+
+	ret = ov5693_write_reg_array(client, ov5693_res[ov5693->fmt_idx].regs);
+	if (ret) {
+		dev_err(&client->dev, "ov5693 write register err.\n");
+		return ret;
+	}
+
+	ret = ov5693_sw_standby(ov5693, true);
+	if (ret)
+		dev_err(&client->dev, "ov5693 stream off error\n");
+
+	return ret;
+}
+
 static void ov5693_sensor_powerdown(struct ov5693_device *ov5693)
 {
 	gpiod_set_value_cansleep(ov5693->reset, 1);
@@ -845,6 +881,12 @@ static int __maybe_unused ov5693_sensor_resume(struct device *dev)
 	ret = ov5693_sensor_powerup(ov5693);
 	if (ret)
 		goto out_unlock;
+
+	ret = ov5693_sensor_init(ov5693);
+	if (ret) {
+		dev_err(&client->dev, "ov5693 sensor init failure\n");
+		goto err_power;
+	}
 
 	if (ov5693->streaming) {
 		ret = ov5693_sw_standby(ov5693, false);
@@ -944,35 +986,6 @@ static int get_resolution_index(int w, int h)
 	return -1;
 }
 
-/* TODO: remove it. */
-static int startup(struct v4l2_subdev *sd)
-{
-	struct ov5693_device *ov5693 = to_ov5693_sensor(sd);
-	struct i2c_client *client = v4l2_get_subdevdata(sd);
-	int ret = 0;
-
-	ret = ov5693_write_reg(client, OV5693_8BIT,
-			       OV5693_SW_RESET, 0x01);
-	if (ret) {
-		dev_err(&client->dev, "ov5693 reset err.\n");
-		return ret;
-	}
-
-	ret = ov5693_write_reg_array(client, ov5693_global_setting);
-	if (ret) {
-		dev_err(&client->dev, "ov5693 write register err.\n");
-		return ret;
-	}
-
-	ret = ov5693_write_reg_array(client, ov5693_res[ov5693->fmt_idx].regs);
-	if (ret) {
-		dev_err(&client->dev, "ov5693 write register err.\n");
-		return ret;
-	}
-
-	return ret;
-}
-
 static int ov5693_set_fmt(struct v4l2_subdev *sd,
 			  struct v4l2_subdev_pad_config *cfg,
 			  struct v4l2_subdev_format *format)
@@ -982,7 +995,6 @@ static int ov5693_set_fmt(struct v4l2_subdev *sd,
 	struct i2c_client *client = v4l2_get_subdevdata(sd);
 	int ret = 0;
 	int idx;
-	int cnt;
 
 	if (format->pad)
 		return -EINVAL;
@@ -1013,34 +1025,6 @@ static int ov5693_set_fmt(struct v4l2_subdev *sd,
 		ret = -EINVAL;
 		goto mutex_unlock;
 	}
-
-	for (cnt = 0; cnt < OV5693_POWER_UP_RETRY_NUM; cnt++) {
-		ret = ov5693_sensor_powerup(ov5693);
-		if (ret) {
-			dev_err(&client->dev, "power up failed\n");
-			continue;
-		}
-
-		ret = startup(sd);
-		if (ret)
-			dev_err(&client->dev, " startup() FAILED!\n");
-		else
-			break;
-	}
-	if (cnt == OV5693_POWER_UP_RETRY_NUM) {
-		dev_err(&client->dev, "power up failed, gave up\n");
-		goto mutex_unlock;
-	}
-
-	/*
-	 * After sensor settings are set to HW, sometimes stream is started.
-	 * This would cause ISP timeout because ISP is not ready to receive
-	 * data yet. So add stop streaming here.
-	 */
-	ret = ov5693_write_reg(client, OV5693_8BIT, OV5693_SW_STREAM,
-			       OV5693_STOP_STREAMING);
-	if (ret)
-		dev_warn(&client->dev, "ov5693 stream off err\n");
 
 mutex_unlock:
 	mutex_unlock(&ov5693->lock);
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -20,19 +20,11 @@
 #include <linux/clk.h>
 #include <linux/delay.h>
 #include <linux/device.h>
-#include <linux/errno.h>
 #include <linux/i2c.h>
-#include <linux/init.h>
-#include <linux/io.h>
-#include <linux/kernel.h>
-#include <linux/kmod.h>
 #include <linux/module.h>
-#include <linux/mm.h>
 #include <linux/pm_runtime.h>
 #include <linux/regulator/consumer.h>
 #include <linux/slab.h>
-#include <linux/string.h>
-#include <linux/types.h>
 #include <media/v4l2-device.h>
 #include <media/v4l2-fwnode.h>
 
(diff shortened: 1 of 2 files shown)


qzed (commented):
Code looks good as far as I can tell. I've noted a small formatting/magic-value thing. Also does the control setup (e.g. `v4l2_ctrl_new_std()`) need checking? If so that'd be something for a later patch to keep in mind.

Testing does show some new errors though (and the screen is black):

- qcam:
  ```
  [0:15:29.437003908] [50840]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+2272-f9c0cf2a
  [0:15:29.447222527] [50842] ERROR V4L2 v4l2_subdevice.cpp:285 'ov5693 2-0036': Unable to get rectangle 2 on pad 0: Invalid argument
  [0:15:29.447352231] [50842]  WARN CameraSensor camera_sensor.cpp:311 'ov5693 2-0036': The PixelArraySize property has been defaulted to (0x0)/2592x1944
  [0:15:29.447371902] [50842]  WARN CameraSensor camera_sensor.cpp:336 'ov5693 2-0036': The sensor kernel driver needs to be fixed
  [0:15:29.447381808] [50842]  WARN CameraSensor camera_sensor.cpp:338 'ov5693 2-0036': See Documentation/sensor_driver_requirements.rst in the libcamera sources for more information
  [0:15:29.452569232] [50842]  INFO IPU3 ipu3.cpp:1038 Registered Camera[0] "\_SB_.PCI0.I2C2.CAMF" connected to CSI-2 receiver 1
  [0:15:29.697317254] [50840]  INFO Camera camera.cpp:890 configuring streams: (0) 1280x720-NV12
  Using software format conversion from NV12
  ```

- dmesg:
  ```
  [   91.379246] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
  [   91.379253] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: incomplete long packet detected
  [   91.379255] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: frame sync error
  ```

drivers/media/i2c/ov5693.c
```diff
@@ -1617,24 +1305,18 @@ static int ov5693_init_controls(struct ov5693_device *ov5693)
...
 	struct v4l2_fwnode_device_properties props;
-	unsigned int i;
 	int ret;
 	int hblank;
 	int vblank_max, vblank_min, vblank_def;
 	int exposure_max;
 
 	ret = v4l2_ctrl_handler_init(&ov5693->ctrl_handler,
-				     ARRAY_SIZE(ov5693_controls));
+				     8);
```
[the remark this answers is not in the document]
djrscally: You're right, that was a bit lazy, my bad. In a one-of-a-kind experience for me this parameter is specifically an _estimate_, so I guess I could do something like OV5693_APPROX_NUM_CTRLS (it works if the answer's wrong, just takes a bit longer).

djrscally:
> Code looks good as far as I can tell. I've noted a small formatting/magic-value thing. Also does the control setup (e.g. `v4l2_ctrl_new_std()`) need checking? If so that'd be something for a later patch to keep in mind.

Yes it does; most drivers skip it, but I agree that that's weird. It needs a check after each one and `v4l2_ctrl_handler_free()` in the failure path; it's on my to-do list

> Testing does show some new errors though (and the screen is black):
> 
>     * qcam:
>       ```
>       [0:15:29.437003908] [50840]  INFO Camera camera_manager.cpp:293 libcamera v0.0.0+2272-f9c0cf2a
>       [0:15:29.447222527] [50842] ERROR V4L2 v4l2_subdevice.cpp:285 'ov5693 2-0036': Unable to get rectangle 2 on pad 0: Invalid argument
>       [0:15:29.447352231] [50842]  WARN CameraSensor camera_sensor.cpp:311 'ov5693 2-0036': The PixelArraySize property has been defaulted to (0x0)/2592x1944
>       [0:15:29.447371902] [50842]  WARN CameraSensor camera_sensor.cpp:336 'ov5693 2-0036': The sensor kernel driver needs to be fixed
>       [0:15:29.447381808] [50842]  WARN CameraSensor camera_sensor.cpp:338 'ov5693 2-0036': See Documentation/sensor_driver_requirements.rst in the libcamera sources for more information
>       [0:15:29.452569232] [50842]  INFO IPU3 ipu3.cpp:1038 Registered Camera[0] "\_SB_.PCI0.I2C2.CAMF" connected to CSI-2 receiver 1
>       [0:15:29.697317254] [50840]  INFO Camera camera.cpp:890 configuring streams: (0) 1280x720-NV12
>       Using software format conversion from NV12
>       ```

These are ok (for now) - more validation added by libcamera that we'll need to account for in the future.

>     * dmesg:
>       ```
>       [   91.379246] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
>       [   91.379253] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: incomplete long packet detected
>       [   91.379255] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: frame sync error
>       ```

_aaaaaarrrrrggghhhhh_ that's infuriating. I spent days getting rid of those errors for me, so it sucks you get them. Any chance you can bisect the series to find the broken commit? It probably will be after either 28ef5ca or 2a32c4f but it would be nice to be sure. 

_Functionally_ I don't think anything has changed; the problem seems to be timing. What's happening (I think) is that the sensor's streaming before the CIO2 is actually ready to receive. Unfortunately since I'm not experiencing the problem it's difficult for me to try to fix. You could try making this change:

```diff
diff --git a/drivers/media/i2c/ov5693.c b/drivers/media/i2c/ov5693.c
index 1b875f75e2f0..bb1e9ce282d9 100644
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -803,6 +803,10 @@ static int ov5693_sensor_powerup(struct ov5693_device *ov5693)
 
        usleep_range(20000, 25000);
 
+       ov5693_sw_standby(ov5693, false);
+       usleep_range(1000, 2000);
+       ov5693_sw_standby(ov5693, true);
+
        return 0;
 
 fail_power:
```

As some drivers include that in the powerup saying it's necessary to ensure the transmitter knows it's in the proper state. Or you could try this:

```diff
diff --git a/drivers/media/i2c/ov5693.h b/drivers/media/i2c/ov5693.h
index 0377853f8b2b..e9cd7599f9b1 100644
--- a/drivers/media/i2c/ov5693.h
+++ b/drivers/media/i2c/ov5693.h
@@ -431,7 +431,7 @@ static struct ov5693_reg const ov5693_global_setting[] = {
        {OV5693_8BIT, 0x4512, 0x01},
        {OV5693_8BIT, 0x4806, 0x00},
        {OV5693_8BIT, 0x4816, 0x52},
-       {OV5693_8BIT, 0x481f, 0x30},
+       {OV5693_8BIT, 0x481f, 0x50},
        {OV5693_8BIT, 0x4826, 0x2c},
        {OV5693_8BIT, 0x4831, 0x64},
        {OV5693_8BIT, 0x4d00, 0x04},
```

Which I _think_ is giving the transmitter more time to settle after power on before sending data. If either of those help maybe we can figure out some values that are good for all of us?

In the meantime, I'm going to ask for some advice in #v4l on this one I think.

qzed:
> _aaaaaarrrrrggghhhhh_ that's infuriating. I spent days getting rid of those errors for me, so it sucks you get them. Any chance you can bisect the series to find the broken commit? It probably will be after either [28ef5ca](https://github.com/linux-surface/kernel/commit/28ef5cac018d4a9d18db983ad460e7a0d8361a2a) or [2a32c4f](https://github.com/linux-surface/kernel/commit/2a32c4f98e9ae29704445ba2b9ef1fd20d191ed1) but it would be nice to be sure.

Sure can do.

qzed:
Okay, so I've been trying around a bit without any changes and the dmesg log seems to change. Sometimes it's the same three messages, sometimes in different order, sometimes only some of them (seem to be either frame sync error alone or the other two together). Doesn't change the black image issue though. Also it shows `0.00 fps`.

I'll try reverting dee27af next, then the two changes, and if that still doesn't work I'll try bisecting.

qzed:
Reverting dee27af doesn't change anything, `dmesg` and `qcam` output is still the same. Both changes you've suggested don't seem to change much either. It feels like there aren't any error messages in `dmesg` the first time when I start `qcam` and maybe there's less output after that, but not as much to be certain.

Guess I'll start bisecting now.

qzed:
Okay... so I started off with the current `ipu3-camers`, which of course didn't work due to missing vblank controls. So I applied that (with minor changes to make that work) directly on top of that and it already stops working with the same errors.

Last time I remember it working was on the v5.10 branch with an older libcamera (before the whole vblanc thing...), so I guess it's caused by changes in libcamera?.

qzed:
Okay, works with libcamera https://github.com/kbingham/libcamera/commit/f1b0c7f5e3f2d55c30934d1f2f9f417e0ff778ae and this full PR. Also the dmesg and libcamera output still appears with that version and the vblank control backported to `ipu3-cameras`. So as far as I can tell there isn't any regression in this PR, so nothing that'd prevent merging this from my side.

Maybe I can find some more time to figure out what exactly is going wrong by bisecting libcamera.

qzed (approved):

djrscally:
> Okay, works with libcamera [kbingham/libcamera@f1b0c7f](https://github.com/kbingham/libcamera/commit/f1b0c7f5e3f2d55c30934d1f2f9f417e0ff778ae) and this full PR. Also the dmesg and libcamera output still appears with that version and the vblank control backported to `ipu3-cameras`. So as far as I can tell there isn't any regression in this PR, so nothing that'd prevent merging this from my side. Guess I just never noticed the dmesg errors... they don't seem to occur on the first qcam run after boot, only on successive runs.
> 
> Maybe I can find some more time to figure out what exactly is going wrong by bisecting libcamera.

You mean these errors still occur running that "make VBLANK mandatory" commit of libcamera and just the first couple of commits from this PR?

```
* dmesg:
  ```
  [   91.379246] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
  [   91.379253] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: incomplete long packet detected
  [   91.379255] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: frame sync error
```

Is that right? If so, could you try running libcamera [e81f904](https://github.com/kbingham/libcamera/commit/f1b0c7f5e3f2d55c30934d1f2f9f417e0ff778ae) without any of this PR at all and see what that does? I wouldn't expect those errors to still show, so that still concerns me. The libcamera output is fine.

qzed:
> > Okay, works with libcamera [kbingham/libcamera@f1b0c7f](https://github.com/kbingham/libcamera/commit/f1b0c7f5e3f2d55c30934d1f2f9f417e0ff778ae) and this full PR. Also the dmesg and libcamera output still appears with that version and the vblank control backported to `ipu3-cameras`. So as far as I can tell there isn't any regression in this PR, so nothing that'd prevent merging this from my side. Guess I just never noticed the dmesg errors... they don't seem to occur on the first qcam run after boot, only on successive runs.
> > Maybe I can find some more time to figure out what exactly is going wrong by bisecting libcamera.
> 
> You mean these errors still occur running that "make VBLANK mandatory" commit of libcamera and just the first couple of commits from this PR?

Actually just the vblank commit, no other commits from this PR (with minor changes to make that work, of course). I've checked and this also happens on the v5.10.13 kernel and on `ipu3-cameras` without this PR and without the VBLANK commit. I think I just didn't notice because this never happens on the first try and I don't really pay much attention to dmesg as long as it does show an image.

> 
> ```
>   [   91.379246] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
>   [   91.379253] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: incomplete long packet detected
>   [   91.379255] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: frame sync error
> ```
> 
> Is that right? If so, could you try running libcamera [e81f904](https://github.com/kbingham/libcamera/commit/e81f9046caa26b06098eaef93b825a08bb3a5564) (parent commit of the one that made VBLANK mandatory) without any of this PR at all and see what that does and then the same libcamera commit but again with the VBLANK added?

Same thing.

> I.E. did you just not notice those errors, or were they not there before. I wouldn't expect those errors to still show, so that still concerns me. The libcamera output is fine.

Yeah, I think I really missed those before... Usually the first try doesn't show anything, the second try sometimes too, but if I repeatedly start and stop qcam, the third or fourth try will show all of those errors. E.g. just now with libcamera  [e81f904](https://github.com/kbingham/libcamera/commit/e81f9046caa26b06098eaef93b825a08bb3a5564) and the vblank commit of this PR backported to ipu3-cameras I get on the first four runs:
1. nothing
2. ```
   [   39.490679] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
   ```
3. ```
   [   44.913066] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: incomplete long packet detected
   [   44.944399] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
   ```
4. ```
   [   47.179448] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: incomplete long packet detected
   [   47.212753] ipu3-cio2 0000:00:14.3: port 1 error PKT2SHORT
   [   47.212758] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: frame sync error
   ```

Could be that I'm breaking something by starting and stopping qcam too quickly.

> It is a bit weird; the last PR didn't include the VBLANK commit and that exhibited these problems too, so maybe it is something from libcamera - though I'm running libcamera master too and I don't get them.

Might be. Again, latest (i.e. https://github.com/kbingham/libcamera/commit/f9c0cf2a3a32ff9306da9931218909fa222e8295) doesn't even show any image for me, so there also seems to be at least another issue in libcamera (or some new control is expected and missing? Although I guess that should break things for you too...).

> Thanks for your efforts btw; those timestamps horrify me though - I thought I stayed up late coding!

Urgh yeah, ever since this whole pandemic thing started, my sleep cycle began to drift. Really need to get that thing back under control...

djrscally:
> Actually just the vblank commit, no other commits from this PR (with minor changes to make that work, of course). I've checked and this also happens on the v5.10.13 kernel and on ipu3-cameras without this PR and without the VBLANK commit. I think I just didn't notice because this never happens on the first try and I don't really pay much attention to dmesg as long as it does show an image.

Right, ok, that's definitely pretty weird then! Guess we need to report that to libcamera so we can look into what's gone crazy. At least it's not this stack of code though, that makes me feel better - thanks for being so thorough!

> Might be. Again, latest (i.e. kbingham/libcamera@f9c0cf2) doesn't even show any image for me, so there also seems to be at least another issue in libcamera (or some new control is expected and missing? Although I guess that should break things for you too...).

It can't be a control because that'd break for me...it feels a lot like some kind of timing issue or a race problem or something. Right pain in the backside to diagnose this one I think.

> Urgh yeah, ever since this whole pandemic thing started, my sleep cycle began to drift. Really need to get that thing back under control...

Heh for me it was the second kid. If I don't stay up I don't have time to do anything!

qzed:
I'll try to bisect libcamera later, that should get us a hint towards one of those problems at least.

djrscally:
> I'll try to bisect libcamera later, that should get us a hint towards one of those problems at least.

Alright; you could also try applying this: https://patchwork.linuxtv.org/project/linux-media/patch/20210215075742.12434-1-sakari.ailus@linux.intel.com/

Chance it has something to do with it I guess

drivers/media/i2c/ov5693.c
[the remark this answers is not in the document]
fabwu: See docs:
```
 * @nr_of_controls_hint: A hint of how many controls this handler is
 *		expected to refer to. This is the total number, so including
 *		any inherited controls. It doesn't have to be precise, but if
 *		it is way off, then you either waste memory (too many buckets
 *		are allocated) or the control lookup becomes slower (not enough
 *		buckets are allocated, so there are more slow list lookups).
 *		It will always work, though.
```
Other drivers use a plain number too so I would leave it as is.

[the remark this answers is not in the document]
qzed: Okay.

drivers/media/i2c/ov5693.c
```diff
@@ -910,6 +916,16 @@ static int ov5693_s_ctrl(struct v4l2_ctrl *ctrl)
 	struct i2c_client *client = v4l2_get_subdevdata(&dev->sd);
 	int ret = 0;
 
+	/* If VBLANK is altered we need to update exposure to compensate */
+	if (ctrl->id == V4L2_CID_VBLANK) {
+		int exposure_max;
+		exposure_max = dev->mode->lines_per_frame - 8; // set 123 for sensible display
```
fabwu: This comment looks like a leftover so remove it.

drivers/media/i2c/ov5693.c
```diff
@@ -1648,10 +1665,10 @@ static int ov5693_init_controls(struct ov5693_device *ov5693)
 	}
 
 	/* Exposure */
-
+	exposure_max = ov5693->mode->lines_per_frame - 8; // Set to 123 for sensible display
```
fabwu, line 1208: This one as well.

drivers/media/i2c/ov5693.h
```diff
@@ -37,68 +37,23 @@
  */
 #define ENABLE_NON_PREVIEW	1
 
-#define OV5693_POWER_UP_RETRY_NUM 5
```
fabwu, line 40: How about merging this header with ov5693.c?

I noticed that newer sensor drivers don't use header files. Maybe to prevent duplication.

fabwu (commented):
Thanks for sharing your changes. VBLANK/HBLANK was also on my list so it's great to see it already done. I try to get ov8865 working before switching back to the ov5693 so we shouldn't interfere with each other. 

@djrscally Btw. what is the current status of the upstream work? 

I added some minor comments and tested the driver:

The driver does work and I can see a very bright pink image although I set the exposure and applied the patch which is a bit weird. I also noticed this error message in `dmesg` with https://github.com/kbingham/libcamera/commit/f9c0cf2:
```
[   29.188439] ov5693 i2c-INT33BE:00: detect ov5693 success
[   43.334069] use of bytesused == 0 is deprecated and will be removed in the future,
[   43.334071] use the actual size instead.
[  216.245119] ipu3-cio2 0000:00:14.3: CSI-2 receiver port 1: frame sync error
```
Maybe this is again some timing issue but I get at least an image back.

Unloading does also work fine but I get this warning:
```
[  875.434540] ov5693 i2c-INT33BE:00: Unbalanced pm_runtime_enable!
```

drivers/media/i2c/ov5693.c
djrscally: Whoops! That was me diagnosing why my image was nuclear explosion white all of a sudden. Thanks, I'll take them out.

drivers/media/i2c/ov5693.h
djrscally: I agree; that's on my list, but still a bit of trimming to do; wanted to get the header as small as possible before merging them...that file is loooooong already.

qzed:
@fabwu Interesting that this works for you, you also have a SB2 right? Anything we might be doing differently?

djrscally:
@fabwu 

> Thanks for sharing your changes. VBLANK/HBLANK was also on my list so it's great to see it already done. I try to get ov8865 working before switching back to the ov5693 so we shouldn't interfere with each other.

Great; I just had a [shifty bash at the ov7251](https://github.com/djrscally/kernel/tree/ov7251) too. No hardware for me though (they have the actual PMIC on the Go2, so I can't power that sensor on) so can't test it myself, but we're probably approaching all 3 cameras "working" I think.

> @djrscally Btw. what is the current status of the upstream work?

The bridge stuff is in linux-next for v5.12 and I've done ~85% of the work on the v3 of intel-skl-int3472; should post that by the weekend I hope. I'm mostly looking at the drivers to have a break from it to be honest.

> I added some minor comments and tested the driver:

Thank you!

> The driver does work and I can see a very bright pink image although I set the exposure and applied the patch which is a bit weird. I also noticed this error message in dmesg with kbingham/libcamera@f9c0cf2:

And yet it doesn't for Maximilian; that's a bit crazy. I have no idea how to diagnose this! Pleased it works for you at least though

> Unloading does also work fine but I get this warning:

Hmmm thanks, let me look into that; shouldn't be happening ofc.

qzed:
Also with v5.11 out, we might want to consider rebasing or opening a new branch and moving to that (whatever is easier) to maybe get rid of some old baggage. Or just directly moving to some (for-)next media branch.

qzed:
Sorry, didn't get around to do more testing yesterday, but I have good news now.

I've just built a new kernel with this PR and latest libcamera completely clean. I've also unset my default `CFLAGS` and `CXXFLAGS` (were both `-pipe -march=native`). After I've noticed that (the apparently optional) `lttng-ust` library was missing I've also installed that and built libcamera with it.

Not entirely sure which part did it, but now it's working as you described. ~2fps, overexposed (largely white), light green tint. No errors in `dmesg`. No clue what went wrong before, seems like that's on me, sorry.

fabwu:
Ok that's weird but I had similar issues before. Sometimes my changes don't show up when I load a module and a `make clean` fixes this problem.

This patch fixes the blinking LED on module load on my Surface Book 2 (yes I have one too):
```diff
diff --git a/drivers/media/i2c/ov5693.c b/drivers/media/i2c/ov5693.c
index 1b875f75e2..57c897dfb3 100644
--- a/drivers/media/i2c/ov5693.c
+++ b/drivers/media/i2c/ov5693.c
@@ -799,7 +799,6 @@ static int ov5693_sensor_powerup(struct ov5693_device *ov5693)
 
        gpiod_set_value_cansleep(ov5693->reset, 0);
        gpiod_set_value_cansleep(ov5693->powerdown, 0);
-       gpiod_set_value_cansleep(ov5693->indicator_led, 1);
 
        usleep_range(20000, 25000);
 
@@ -845,6 +844,8 @@ static int __maybe_unused ov5693_sensor_resume(struct device *dev)
        if (ret)
                goto out_unlock;
 
+       gpiod_set_value_cansleep(ov5693->indicator_led, 1);
+
        ret = ov5693_sensor_init(ov5693);
        if (ret) {
                dev_err(&client->dev, "ov5693 sensor init failure\n");
@@ -1273,7 +1274,7 @@ static int ov5693_configure_gpios(struct ov5693_device *ov5693)
         }
 
         ov5693->indicator_led = devm_gpiod_get_optional(&ov5693->client->dev, "indicator-led",
-                                        GPIOD_OUT_HIGH);
+                                        GPIOD_OUT_LOW);
         if (IS_ERR(ov5693->indicator_led)) {
                 dev_err(&ov5693->client->dev, "Couldn't find indicator-led GPIO\n");
                 return PTR_ERR(ov5693->indicator_led);
```
I'm not sure if `sensor_resume` is the correct place to enable the indicator LED but at least it isn't called during probing. Additionally, I had to switch form `HIGH` to `LOW`.

@qzed @djrscally Does this patch resolve the blinking LED problem on your device?

qzed:
@fabwu Can confirm that your patch works!

drivers/media/i2c/ov5693.c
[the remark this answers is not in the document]
djrscally: Fixed the formatting, leaving the magic number :)

drivers/media/i2c/ov5693.c
djrscally: Removed both these - thanks.

djrscally:
Fixed those comments that @fabwu noted and the formatting that @qzed noted.

> Sorry, didn't get around to do more testing yesterday, but I have good news now.
> 
> I've just built a new kernel with this PR and latest libcamera completely clean. I've also unset my default `CFLAGS` and `CXXFLAGS` (were both `-pipe -march=native`). After I've noticed that (the apparently optional) `lttng-ust` library was missing I've also installed that and built libcamera with it.
> 
> Not entirely sure which part did it, but now it's working as you described. ~2fps, overexposed (largely white), light green tint. No errors in `dmesg`. No clue what went wrong before, seems like that's on me, sorry.

Now that is weird yeah...@fabwu any chance of you trying the same thing? I'd love to get them all cleared before pressing merge, although I think this is ok without it.

> @qzed @djrscally Does this patch resolve the blinking LED problem on your device?

Ah! Nice one, thanks. That fixes it for me too...although in the v3 of intel-skl-int3472 that GPIO line will be turned on when the clock turns on; the linux folks weren't happy with drivers directly controlling the "is this thing on" LED.

> Also with v5.11 out, we might want to consider rebasing or opening a new branch and moving to that (whatever is easier) to maybe get rid of some old baggage. Or just directly moving to some (for-)next media branch.

Yeah sure, that's a good idea. I did just open #83, but I'll not open any more till we get everything squared away then and we can open a new branch.

qzed:
> Now that is weird yeah...@fabwu any chance of you trying the same thing? I'd love to get them all cleared before pressing merge, although I think this is ok without it.

I'm still trying to wrap my head around what went wrong there... I'm pretty sure the kernel build was fine as rebuilding that without the vblank commit caused libcamera to complain. On the other hand, rolling back the libcamera commit got that working again too... I've also failed to reproduce this by setting the C(XX)FLAGS again. Seeing as fully cleaning everything fixed it, I'm inclined to write this off as one of those build/packaging miracles and not worry too much about it.

fabwu:
> I'm inclined to write this off as one of those one-time build/packaging miracles and not worry too much about it.

I think it's something along that line. I had similar issues before so I think we can merge this thing.

@qzed I saw that you included enabled the ov8865 in the 5.11 build. This driver is not quite ready and will throw an error so you better disable it.

qzed:
> @qzed I saw that you included enabled the ov8865 in the 5.11 build. This driver is not quite ready and will throw an error so you better disable it.

Hmm, I thought I didn't. I've pulled a couple of newer commits but I didn't include 4f3d253648d3cacabc291ee0102b076e66b1aa3f, d37bf622b7fa98caee1e78e5dfdb0927bf52e89f, and 6bb1f73dea1464eae9d34c8454ce793421606da3. Anything else I should have excluded?

fabwu:
@djrscally I'm a bit confused about hblank/vblank again. In the ov7251 driver the hblank calculation is quite simple:
```
hblank = ov7251->current_mode->hts - ov7251->current_mode->width;
```
whereas in the ov8865 (very nice driver) the calculation is like this:
```
h_blank = to_pixels_per_line(ov8856->cur_mode->hts, ov8856->cur_mode->link_freq_index) - ov8856->cur_mode->width;

static u64 to_pixels_per_line(u32 hts, u32 f_index)
{
	u64 ppl = hts * to_pixel_rate(f_index);

	do_div(ppl, OV8856_SCLK);

	return ppl;
}
```
Do you know what's the difference between the two?

djrscally:
> @djrscally I'm a bit confused about hblank/vblank again. In the ov7251 driver the hblank calculation is quite simple:
> 
> ```
> hblank = ov7251->current_mode->hts - ov7251->current_mode->width;
> ```
> 
> whereas in the ov8865 (very nice driver) the calculation is like this:
> 
> ```
> h_blank = to_pixels_per_line(ov8856->cur_mode->hts, ov8856->cur_mode->link_freq_index) - ov8856->cur_mode->width;
> 
> static u64 to_pixels_per_line(u32 hts, u32 f_index)
> {
> 	u64 ppl = hts * to_pixel_rate(f_index);
> 
> 	do_div(ppl, OV8856_SCLK);
> 
> 	return ppl;
> }
> ```
> 
> Do you know what's the difference between the two?

So my understanding is this (could very easily be wrong, it's a bit of a mind****) most of the time, pixels per line and HTS are the same value, but that's because the drivers are almost all hardcoding the frequencies and so they can just define the value and leave it. Paul in his quest for adaptability seems to be defining a base value that the actual pixels per line can be calculated from (because it depends on the frequency). The formula for pixels per line is `(HTS / SCLK) * PIXEL_RATE`, but usually SCLK and PIXEL_RATE are hard-coded as the same value so it reduces to just be HTS.

In the ov8865, the link frequency is 360MHz:

```c
static const s64 ov8865_link_freq_menu[] = {
	360000000,
};
```

Which means the pixel rate is 360000000 * 2 * nr_lanes / bits-per-sample. He's also allowing for different numbers of lanes in the driver rather than hardcoding that...but assuming 2 lanes it's a pixel-rate of 144MHz. He provides two possible configurations for PLL2. In the first, SCLK works out as 144MHz also so the pixels-per-line == HTS holds true. In the second he's dropping the SCLK to 80MHz which means the pixels per line is somewhat higher than HTS. Have to figure out why

In the ov7251; it's a 1 lane chip with a 240MHz link frequency. That makes pixel rate 48MHz; the existing 24MHz input clock was configuring SCLK to be 48MHz, I matched that for the 19.2MHz input clock config, so `((HTS / SCLK) * PIXEL_RATE) == HTS` - safe to hardcode it.

fabwu:
Thanks! The numbers match up so your explanation makes sense.

PR #90, opened 2021-03-27
Enable wifi powersaving by default again

jonas2515:
The wifi firmware crashes are now fixed, so let's try reenabling wifi powersaving again, it's really important for power consumption and we should really try keeping it enabled.

Since @kitakar5525 mentioned in https://github.com/linux-surface/kernel/pull/70#issuecomment-726087363 that this causes connection issues on the SB1 and S3, I'd prefer if we could test this MR at least on one of those two devices before merging.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/cfg80211.c
+++ b/drivers/net/wireless/marvell/mwifiex/cfg80211.c
@@ -451,13 +451,6 @@ mwifiex_cfg80211_set_power_mgmt(struct wiphy *wiphy,
 		return -EPERM;
 	}
 
-	if (ps_mode)
-		mwifiex_dbg(priv->adapter, MSG,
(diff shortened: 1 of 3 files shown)


qzed (approved):
Both this and #89 have "mwifiex: Use non-posted PCI register writes", but I think git should be able to handle that. Otherwise I'll fix that when merging.

jonas2515:
I'm not sure if it can handle it actually, I'll just update this branch again after #70 is merged.

qzed:
Command-line merge worked without problems, I think GitHub should indicate any issues once #70 is merged, so we'll see then.

jonas2515:
This should go in together with https://github.com/linux-surface/kernel/pull/91, which makes waking up the firmware from powersaving states more reliable.

PR #91, opened 2021-03-28
mwifiex: Try waking the firmware until we get an interrupt

jonas2515:
It seems that the firmware of the 88W8897 card sometimes ignores or
misses when we try to wake it up by reading the firmware status
register. This leads to the firmware wakeup timeout expiring and the
driver resetting the card because we assume the firmware has hung up or
crashed (unfortunately that's not unlikely with this card).

Turns out that most of the time the firmware actually didn't hang up,
but simply "missed" our wakeup request and doesn't send us an AWAKE
event.

Trying again to read the firmware status register after a short timeout
usually makes the firmware wake we up as expected, so add a small retry
loop to mwifiex_pm_wakeup_card() that looks at the interrupt status to
check whether the card woke up.

The number of tries and timeout lengths for this were determined
experimentally: The firmware usually takes about 500 us to wake up
after we attempt to read the status register. In some cases where the
firmware is very busy (for example while doing a bluetooth scan) it
might even miss our requests for multiple milliseconds, which is why
after 15 tries the waiting time gets increased to 10 ms. The maximum
number of tries it took to wake the firmware when testing this was
around 20, so a maximum number of 50 tries should give us plenty of
safety margin.

A good reproducer for this issue is letting the firmware sleep and wake
up in very short intervals, for example by pinging an device on the
network every 0.1 seconds.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/pcie.c
+++ b/drivers/net/wireless/marvell/mwifiex/pcie.c
@@ -665,6 +665,7 @@ static int mwifiex_pm_wakeup_card(struct mwifiex_adapter *adapter)
 {
 	struct pcie_service_card *card = adapter->card;
 	const struct mwifiex_pcie_card_reg *reg = card->pcie.reg;
+	int n_tries = 0;
 
 	mwifiex_dbg(adapter, EVENT,
 		    "event: Wakeup device...\n");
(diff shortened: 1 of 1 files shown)


drivers/net/wireless/marvell/mwifiex/pcie.c
```diff
@@ -665,20 +665,50 @@ static int mwifiex_pm_wakeup_card(struct mwifiex_adapter *adapter)
...
 	}
 
+	n_tries = 1;
+
+	usleep_range(400, 700);
+
+	/* adapter->int_status is a single byte, so should be fine to access
+	 * without aquiring the spin-lock.
+	 */
+	while (n_tries < 50 && adapter->int_status == 0) {
```
qzed, line 694: The `adapter->int_status` should probably be wrapped with `READ_ONCE` to prevent the compiler from optimizing this check away somehow, i.e. `READ_ONCE(adapter->int_status)`

drivers/net/wireless/marvell/mwifiex/pcie.c
```diff
@@ -665,20 +665,50 @@ static int mwifiex_pm_wakeup_card(struct mwifiex_adapter *adapter)
...
+			return -1;
+		}
+
+		if (n_tries < 15)
+			usleep_range(400, 700);
+		else
+			msleep(10);
+
+		n_tries++;
+	}
```
qzed, line 707: I think you could move all of this into a single do-while loop. Initialize `n_tries` above to zero and then

```c
do {
	if (n_tries >= 15)
		msleep(10);
	else if (n_tries)
		usleep_range(400, 700);

	if (mwifiex_write_reg(adapter, reg->fw_status, FIRMWARE_READY_PCIE)) {
		mwifiex_dbg(adapter, ERROR, "Writing fw_status register failed\n");
		return -1;
	}

	n_tries++;
} while (n_tries < 50 && adapter->int_status == 0)
```

would make this a bit more compact.

qzed (changes requested):
One thing where compiler optimization could screw us a bit and some formatting where you could reduce code a bit.

drivers/net/wireless/marvell/mwifiex/pcie.c
jonas2515: Did that

drivers/net/wireless/marvell/mwifiex/pcie.c
jonas2515: Good idea! Did something similar to this

qzed (approved):

PR #92, opened 2021-04-13
mwifiex: Fix host sleep without powersaving and add quirk to disable deep sleep with bad hw revisions

jonas2515:
The 88W8897 pcie card with the hardware revision 20 apparently has a
hardware issue where the card wakes up from deep sleep randomly and very
often, somewhat depending on the card activity, maybe the hardware has a
floating wakeup pin or something.

Those continuous wakeups prevent the card from entering host sleep when
the computer suspends. And because the host won't answer to events from
the card anymore while it's suspended, the firmwares internal
powersaving state machine seems to get confused and the card can't sleep
anymore at all after that.

Since we can't work around that hardware bug in the firmware, let's
get the hardware revision string from the firmware and match it with
known bad revisions. Then disable auto deep sleep for those revisions,
which makes sure we no longer get those spurious wakeups.

### DIFF
--- a/drivers/net/wireless/marvell/mwifiex/main.c
+++ b/drivers/net/wireless/marvell/mwifiex/main.c
@@ -401,6 +401,12 @@ process_start:
 		     !adapter->scan_processing) &&
 		    !adapter->data_sent &&
 		    !skb_queue_empty(&adapter->tx_data_q)) {
+			if (adapter->hs_activated_manually) {
+				mwifiex_cancel_hs(mwifiex_get_priv(adapter, MWIFIEX_BSS_ROLE_ANY),
+						  MWIFIEX_ASYNC_CMD);
+				adapter->hs_activated_manually = false;
+			}
+
 			mwifiex_process_tx_queue(adapter);
 			if (adapter->hs_activated) {
 				clear_bit(MWIFIEX_IS_HS_CONFIGURED,
@@ -418,6 +424,12 @@ process_start:
 		    !mwifiex_bypass_txlist_empty(adapter) &&
 		    !mwifiex_is_tdls_chan_switching
 			(mwifiex_get_priv(adapter, MWIFIEX_BSS_ROLE_STA))) {
+			if (adapter->hs_activated_manually) {
+				mwifiex_cancel_hs(mwifiex_get_priv(adapter, MWIFIEX_BSS_ROLE_ANY),
+						  MWIFIEX_ASYNC_CMD);
+				adapter->hs_activated_manually = false;
+			}
+
 			mwifiex_process_bypass_tx(adapter);
 			if (adapter->hs_activated) {
 				clear_bit(MWIFIEX_IS_HS_CONFIGURED,
@@ -434,6 +446,12 @@ process_start:
 		    !adapter->data_sent && !mwifiex_wmm_lists_empty(adapter) &&
 		    !mwifiex_is_tdls_chan_switching
 			(mwifiex_get_priv(adapter, MWIFIEX_BSS_ROLE_STA))) {
+			if (adapter->hs_activated_manually) {
+				mwifiex_cancel_hs(mwifiex_get_priv(adapter, MWIFIEX_BSS_ROLE_ANY),
+						  MWIFIEX_ASYNC_CMD);
+				adapter->hs_activated_manually = false;
+			}
+
 			mwifiex_wmm_process_tx(adapter);
 			if (adapter->hs_activated) {
 				clear_bit(MWIFIEX_IS_HS_CONFIGURED,
--- a/drivers/net/wireless/marvell/mwifiex/main.c
+++ b/drivers/net/wireless/marvell/mwifiex/main.c
@@ -226,6 +226,19 @@ exit_rx_proc:
 	return 0;
 }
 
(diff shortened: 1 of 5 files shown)


drivers/net/wireless/marvell/mwifiex/main.c
```diff
@@ -226,6 +226,19 @@ static int mwifiex_process_rx(struct mwifiex_adapter *adapter)
...
+{
+	struct mwifiex_private *priv = mwifiex_get_priv(adapter, MWIFIEX_BSS_ROLE_STA);
+	struct mwifiex_ver_ext ver_ext;
+
+	memset(&ver_ext, 0, sizeof(ver_ext));
+	ver_ext.version_str_sel = 1;
+	mwifiex_send_cmd(priv, HostCmd_CMD_VERSION_EXT,
+			 HostCmd_ACT_GEN_GET, 0, &ver_ext, false);
+
+	priv->ds_quirk_version_string_requested = true;
```
qzed, line 239: Is it correct that you're setting this _after_ sending the command or could the command be theoretically completed in `mwifiex_ret_ver_ext()` before you set this to `true`?

drivers/net/wireless/marvell/mwifiex/sta_cmdresp.c
```diff
@@ -708,6 +708,23 @@ static int mwifiex_ret_ver_ext(struct mwifiex_private *priv,
...
+			mwifiex_dbg(priv->adapter, MSG,
+				    "Bad HW revision detected, disabling deep sleep\n");
+
+			mwifiex_send_cmd(priv, HostCmd_CMD_802_11_PS_MODE_ENH,
+					 DIS_AUTO_PS, BITMAP_AUTO_DS, &auto_ds, false);
+		}
+
+		priv->ds_quirk_version_string_requested = false;
+		return 0;
+	}
```
qzed, line 726: Does this need some sort of synchronization? If so, maybe replace that with atomic bitops, e.g. `set_bit()` in `maybe_quirk_fw_disable_ds()` and `test_and_clear_bit()` here.

I guess this should work as is though as `mwifiex_ret_ver_ext()` and `maybe_quirk_fw_disable_ds()` both shouldn't be entered multiple times in parallel?

qzed (commented):
Just tested this, and this fixes the suspend issue.

Two small comments regarding possible synchronization issues below. Looks good otherwise.

drivers/net/wireless/marvell/mwifiex/main.c
jonas2515: For clarity it might be a good idea to set `ds_quirk_version_string_requested` to `true` before sending the command, but to answer the question: No, we won't invoke into the command response handler synchronously, because `false` was passed as the `sync` argument of `mwifiex_send_cmd()`, which means the command gets pushed to the end of the command queue and will be executed later.

drivers/net/wireless/marvell/mwifiex/sta_cmdresp.c
jonas2515: I think the only risk here is that a user-invoked (from sysfs) verext command gets queued right before we queue ours. Then the response handler for the user-invoked command will get called first and `ds_quirk_version_string_requested` will be set to `false`. Not sure if we want to care about that..

But yeah, using `set_bit()` and `test_and_clear_bit()` sounds good indeed.

drivers/net/wireless/marvell/mwifiex/main.c
qzed: But is there anything that'd prevent a scenario like:
- Code executes until right after `mwifiex_send_cmd()`.
- Kernel decides to un-schedule the current thread before `priv->ds_quirk_version_string_requested = true;` and schedules the (probably workqueue) thread processing the command that just got queued.
- Command gets executed and completed before the verext thread gets re-scheduled and the flag gets set to true.

Stuff like that is probably very unlikely due to the time it takes to talk to the firmware, but theoretically still possible, unless there's some other synchronization going on (like this code being executed on the same thread that processes the command workqueue).

drivers/net/wireless/marvell/mwifiex/sta_cmdresp.c
qzed: Right, I think with `set_bit()` and `test_and_clear_bit()` you don't have to think about any such stuff and it should "just work".

drivers/net/wireless/marvell/mwifiex/main.c
jonas2515: Oh yes, this code is being executed in the workqueue thread, I didn't mention that. That's actually exactly the reason why we can't run the command synchronously, because blocking while inside the loop that executes commands and gets responses is obviously not going to work.

drivers/net/wireless/marvell/mwifiex/sta_cmdresp.c
jonas2515: Alright, switched to using a bit now

drivers/net/wireless/marvell/mwifiex/main.c
qzed: Got it, then that will work.

qzed (approved):

PR #94, opened 2021-04-21
mwifiex: Ignore BTCOEX events from the firmware

jonas2515:
The firmware of the pcie 88W8897 chip sends those events very
unreliably, which means we sometimes end up actually capping the window
size while bluetooth is disabled, artifically limiting wifi speeds even
though it's not needed.

Since we can't fix the firmware, let's just ignore those events, it
seems that the Windows driver also doesn't change the rx/tx block ack
buffer sizes when bluetooth gets enabled or disabled, so this is
consistent with the Windows driver.


qzed (approved):
Thanks! Looks good to me, as far as I can tell anyways.

PR #95, opened 2021-05-06
cio2-bridge: Use correct dev_properties size

fabwu:
Orientation adds another entry to dev_properties. Increase array size accordingly.


djrscally (approved):
Yep, this is right.

qzed (approved):
Thanks!

PR #96, opened 2021-05-11
Fixes and improved support for AMD version of Surface Laptop 4

qzed:
Add device IDs required for GPIO controller and UART on the SL4, as well as one fix/workaround required for its legacy PIC.


qzed:
Possible thanks to @nakato.

PR #97, opened 2021-05-20
Camera Updates for v5.12

qzed:
I still need to test this. Cc @djrscally, if you want to have a look over this.


qzed:
Oh, also Cc @fabwu.

drivers/media/i2c/ov5693.c
```diff
@@ -0,0 +1,1557 @@
...
+#define OV5693_OFFSET_START_Y_H_REG		0x3812
+#define OV5693_OFFSET_START_Y_H(v)		(((v) & GENMASK(15, 8)) >> 8)
+#define OV5693_OFFSET_START_Y_L_REG		0x3813
+#define OV5693_OFFSET_START_Y_L(v)		((v) & GENMASK(7, 0))
+
+#define OV5693_SUB_INC_X_REG			0x3814
+#define OV5693_SUB_INC_Y_REG			0x3815
+
+#define OV5693_FORMAT1_REG			0x3820
+#define OV5693_FORMAT1_FLIP_VERT_ISP_EN		BIT(2)
```
djrscally: We should make this `BIT(6)` and then that will fix the purple image that the folks with a rotated sensor are experiencing. Do you want to do that, or should I drop a patch on top?

djrscally:
Tested and working on both my devices. Still the CIO2 errors; we eventually tracked those down to the power on / off sequence for the ov5693 driver but haven't managed to MacGyver a fix yet

drivers/media/i2c/ov5693.c
qzed: Ah neat, feel free to push one to this branch.

fabwu:
Does work on my SB2. I get around 22fps not sure if I should get more but the image looks ok.

The upstream rotation patch should make progress as soon as a new acpica version is published.

qzed:
Thanks guys! Picture looks good on my SB2 (no more purple, yay), I get about 30fps. Maybe a bit on the blocky side when I maximize the window. Apart from that I think the only other remaining nits are the error messages (although they don't seem to have any impact on it working) and the led flashing once during boot. So pretty good progress so far!

PR #98, opened 2021-05-29
Override irq7 for SL4A over probing Legacy PIC

nakato:
From the mailing list [thread](https://lore.kernel.org/lkml/20210512210459.1983026-1-luzmaximilian@gmail.com/) it sounds like using a quirk to provide an override to the irq 7 interrupt for setup of the IOAPIC is a cleaner hack for the short term.

The two commits here revert the legacy PIC workaround, and copies Thomas Gleixner's debugging patch as a temporary solution.


qzed (approved):
Thanks! I'm kinda wondering if we need to override other IRQs too. We'll probably have a better chance figuring this out with the new workaround.

PR #99, opened 2021-05-29
platform/x86: amd-pmc: Add device HID for AMD PMC

nakato:
The Surface Laptop 4 appears to have used AMD0005 for the PMC instead of
the AMDI0005 which would match the ACPI ID Registry.

AMD appears to have previously used "AMD" in a number of IDs in the past,
and AMD is not allocated to any other entity as an ID, so adding this ID
should not cause any harm.

Signed-off-by: Sachi King <nakato@nakato.io>

### DIFF
--- a/drivers/platform/x86/amd-pmc.c
+++ b/drivers/platform/x86/amd-pmc.c
@@ -275,6 +275,7 @@ static int amd_pmc_remove(struct platform_device *pdev)
 static const struct acpi_device_id amd_pmc_acpi_ids[] = {
 	{"AMDI0005", 0},
 	{"AMD0004", 0},
+	{"AMD0005", 0},
 	{ }
 };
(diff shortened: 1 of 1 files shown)


nakato:
This seems to be part of the puzzle with the device and being able to suspend reliably.
The device does not expose the S3 state, which means it has to rely on S0ix sleep states, and as far as I can tell the amd-pmc is quite important to that.

Now that I know the amd-pmc is present, I can go pay attention and actually read some issues in the drm/amd freedesktop gitlab.

[SL4A](https://github.com/linux-surface/acpidumps/blob/master/surface_laptop_4_amd/ssdt3.dsl#L27)
[SL3A](https://github.com/linux-surface/acpidumps/blob/4da0148744164cea0c924dab92f45842fde03177/surface_laptop_3_amd/ssdt9.dsl)

qzed (approved):
Thanks!

PR #100, opened 2021-06-03
Added quirk for Surface Go touchscreen

zoltantamasvajda:
The Elantech touchscreen/digitizer in the Surface Go mistakenly shows up as having capable of reporting battery (even though it actually cannot). This results in a "low battery" message popping up every time you try to use the pen. This patch adds a quirk allowing the kernel to ignore the the non-existent battery and get rid of the false low battery messages.


qzed (approved):
Thanks!

qzed:
Do you plan on taking this upstream?

zoltantamasvajda:
I can try and see if they want it. This is more of a workaround than a real solution though. I think the reason why it reports having a battery is that the device does not get identified properly so it uses the default values for most properties which just so happens to be true for having a battery. Adding it as a quirk is just kicking the can down the road until someone makes a custom driver.

qzed:
Given that there isn't any driver like this in the works yet (AFAIK), it's probably a good idea to have that upstream. It can always be removed later on. Also there seem to be at least two other workarounds of this sort (https://github.com/linux-surface/kernel/commit/7c38e769d5c508939ce5dc26df72602f3c902342, https://github.com/linux-surface/kernel/commit/decfe496fe77061dea658a0bfa11afd4f92b540d), so this might be a more or less common issue.

zoltantamasvajda:
Patch submitted upstream. I'll let you know how it goes.

zoltantamasvajda:
Patch applied upstream! 🥳

qzed:
Awesome, thanks!

PR #104, opened 2021-09-18
Amd pinctrl wakeup

nakato:
These patches are staged for inclusion in upstream, and they are required
for interrupts that are generated from GPIO pins to be triggered, such
as the Surface Laptop 4's LID switch.

This should also hold true for the Surface Laptop 3's lid.

I pulled these patches from the patch series at:
https://patchwork.ozlabs.org/project/linux-gpio/cover/20210831120613.1514899-1-Basavaraj.Natikar@amd.com/

Ref:
AMD CPU: Waking from suspend doesn't work without external keyboard
https://github.com/linux-surface/linux-surface/issues/80
[SL4A] S0ix support on Surface Laptop 4 AMD
https://github.com/linux-surface/linux-surface/issues/458


qzed (approved):

qzed:
Thanks!

PR #105, opened 2021-09-19
platform/surface: aggregator: reconfigure GPIO interrupt on probe

nakato:
On system initialization the IRQ may be preconfigured by gpiolib-acpi as
per the settings defined in the machine's ACPI table.  To deconfigure
this interrupt as being set as a wake source for Si0x and S3 sleep
states we need to call disable_irq_wake, however there is some error
checking here to ensure the calls are balanced, so we enable the wakeup
source first, before following up with disabling.

Signed-off-by: Sachi King <nakato@nakato.io>

### DIFF
--- a/drivers/platform/surface/aggregator/controller.c
+++ b/drivers/platform/surface/aggregator/controller.c
@@ -2705,6 +2705,21 @@ int ssam_irq_setup(struct ssam_controller *ctrl)
 	if (irq < 0)
 		return irq;
 
+	/*
+	 * GPIO interrupts declared in ACPI are setup for us by the ACPI
+	 * driver before we configure the IRQ, and thus may already be
+	 * enabled.  This driver however does not currently implement wakeup
+	 * detection, so we need the opportunity to reconfigure the configured
+	 * GPIO interrupt, otherwise we will receive spurious interrupts.
+	 * We are not allowed to have unbalanced enable/disable of IRQ wake,
+	 * so we will enable the wakeup source so it can be reconfigured with
+	 * the subsequent disable.
+	 */
+	status = enable_irq_wake(irq);
+	if (status)
+		return status;
+	disable_irq_wake(irq);
+
 	status = request_threaded_irq(irq, NULL, ssam_irq_handle, irqf,
 				      "ssam_wakeup", ctrl);
 	if (status)


nakato:
The GPIO could probably also be ignored via a quirk in gpiolib-acpi, but it isn't a hardware issue, so I don't think we should handle it there.  It would also be troublesome to maintain and possibly require constant rediscovery for new Surface devices.

Prior to this the GPIO is configured as such on boot.
```
pin7    Edge trigger| Active high| interrupt is enabled| interrupt is masked| enable wakeup in S0i3 state| enable wakeup in S3 state|
```

If the machine is suspended and resumed, via a working wakeup source, this GPIO will remain in this configuration.  If anything pokes GPIO 7 while suspended, power shoots up to 10 watts.  (Loaded to a TTY my meter reports 6 watt)



If console0-0 is configured as a wakeup source, and the machine is suspended, then resumed, the pin will now be configured as such.
```
pin7    Edge trigger| Active high| interrupt is enabled| interrupt is masked| disable wakeup in S0i3 state| disable wakeup in S3 state|
```

If console0-0 as a wakeup source is disabled at this point, the gpio interrupt will not be reconfigured to enable wakeup, and the machine will stay asleep and in a low power state.

By toggling the IRQ as a wake source on, [for balance](https://elixir.bootlin.com/linux/latest/source/kernel/irq/manage.c#L901), then off again, we successfully take over the configuration of the GPIO pin from pre-configured ACPI values.

qzed:
This seems a bit like a hack, are you sure about this?

The check for balanced calls is deliberate. That means that the default for IRQs must be disabled as you can otherwise not disable them (as in this instance) without having that check trigger (I'm not sure how this translates to the underlying GPIO pins, however).

There's https://www.kernel.org/doc/Documentation/power/suspend-and-interrupts.rst for more info, specifically:

> Device interrupt request lines (IRQs) are generally disabled during system suspend after the "late" phase of suspending devices (that is, after all of the ->prepare, ->suspend and ->suspend_late callbacks have been executed for all devices).  That is done by suspend_device_irqs().
>
> The rationale for doing so is that after the "late" phase of device suspend there is no legitimate reason why any interrupts from suspended devices should trigger and if any devices have not been suspended properly yet, it is better to block interrupts from them anyway.  Also, in the past we had problems with interrupt handlers for shared IRQs that device drivers implementing them were not prepared for interrupts triggering after their devices had been suspended. In some cases they would attempt to access, for example, memory address spaces of suspended devices and cause unpredictable behavior to ensue as a result.
Unfortunately, such problems are very difficult to debug and the introduction of suspend_device_irqs(), along with the "noirq" phase of device suspend and resume, was the only practical way to mitigate them.


> System wakeup interrupts generally need to be configured to wake up the system from sleep states, especially if they are used for different purposes (e.g. as I/O interrupts) in the working state.
>
> [...]
>
> The IRQ subsystem provides two helper functions to be used by device drivers for those purposes.  Namely, enable_irq_wake() turns on the platform's logic for handling the given IRQ as a system wakeup interrupt line and disable_irq_wake() turns that logic off.

Also note that the two calls don't do anything if the IRQ has already been marked for wake before via `enable_irq_wake()` due to the resource-counting mechanism. So if this really changes behavior, that means that the IRQ hasn't been marked as wake via the official way yet. And If I'm interpreting the above correctly, it's generally expected that IRQs are non-wake until configured as such (again, I don't see any method to mark IRQs as non-wake without first having marked them as wake).

Furthermore, the definition of GPIO interrupt on the [SB2](https://github.com/linux-surface/acpidumps/blob/f8db3d150815aa21530635b7e646eee271e3b8fe/surface_book_2/dsdt.dsl#L15274-L15279) isn't different than the one from the [AMD SL4](https://github.com/linux-surface/acpidumps/blob/f8db3d150815aa21530635b7e646eee271e3b8fe/surface_laptop_4_amd/dsdt.dsl#L2118-L2123). Both are marked as `SharedAndWake`. So at the very least there's some difference between the AMD and Intel pinctrl drivers.

I can't really see where the `WAKE_CNTRL_OFF_S0I3` bit is set other than `amd_gpio_irq_set_wake()`, which is only used via `irq_set_irq_wake()` (and thus I think only `enable_irq_wake()`/`disable_irq_wake()`, which we've established couldn't have been called before). I think since that's a memory register, the default value probably comes from hardware. So to me this seems like Linux by default expects all IRQs to be non-wake, while the driver just pulls its defaults from the hardware, which in this instance has it configured as wake. Also I'm not sure just how much other configuration goes into the `enable_irq_wake()` call, so I think it's entirely possible that there's some sort of state misconfiguration where the hardware thinks it's on and the kernel thinks it's off because no `enable_irq_wake()` call has happened.

So to me it seems that this holds for all interrupts marked as `WAKE_CNTRL_OFF_S0I3` by hardware, not just the SAM one. If I'm understanding all of this right, the proper solution would be to disable all wake pins at boot and then let `enable_irq_wake()` take care of it as (I believe) is expected.

As a note: The ACPI GPIO handler (which I believe is used for the lid) also [explicitly calls `enable_irq_wake()`]( https://elixir.bootlin.com/linux/latest/source/drivers/gpio/gpiolib-acpi.c#L237). So that should still work after initializing all pins to non-wake.

I think it'd be good if we had someone of the pinctrl/GPIO team look at this.

nakato:
> This seems a bit like a hack

Oh, this is absolutely a hack.

> are you sure about this?

Not anymore.

Originally I thought the state of the GPIO Int was being setup elsewhere, and the problem was that it is supposed to be a wake interrupt and we're attempting to not use it as such.  This is incorrect, and there's other wakeups enabled that shouldn't be on top of this one.

> The check for balanced calls is deliberate. That means that the default for IRQs must be disabled as you can otherwise not disable them (as in this instance) without having that check trigger (I'm not sure how this translates to the underlying GPIO pins, however).
> [...]
> Also note that the two calls don't do anything if the IRQ has already been marked for wake before via `enable_irq_wake()` due to the resource-counting mechanism. So if this really changes behavior, that means that the IRQ hasn't been marked as wake via the official way yet. And If I'm interpreting the above correctly, it's generally expected that IRQs are non-wake until configured as such (again, I don't see any method to mark IRQs as non-wake without first having marked them as wake).

You're right, enable hasn't been called on it.  When I was looking at gpiolib-acpi, I somehow thought that GPIO Pin 0x0007 pin was getting setup in there, but I clearly missed the detail that pins that are configured in gpiolib-acpi are only GPIO Interrupt's that are described in an `_AEI` section.

> Furthermore, the definition of GPIO interrupt on the [SB2](https://github.com/linux-surface/acpidumps/blob/f8db3d150815aa21530635b7e646eee271e3b8fe/surface_book_2/dsdt.dsl#L15274-L15279) isn't different than the one from the [AMD SL4](https://github.com/linux-surface/acpidumps/blob/f8db3d150815aa21530635b7e646eee271e3b8fe/surface_laptop_4_amd/dsdt.dsl#L2118-L2123). Both are marked as `SharedAndWake`. So at the very least there's some difference between the AMD and Intel pinctrl drivers.
>
> I think since that's a memory register, the default value probably comes from hardware.

Yep, the /sys/debug/kernel/gpio output is generated directly via reading from the chip.
I suspect this must be preconfigured by the firmware.

> So to me this seems like Linux by default expects all IRQs to be non-wake, while the driver just pulls its defaults from the hardware, which in this instance has it configured as wake. Also I'm not sure just how much other configuration goes into the `enable_irq_wake()` call, so I think it's entirely possible that there's some sort of state misconfiguration where the hardware thinks it's on and the kernel thinks it's off because no `enable_irq_wake()` call has happened.

I had taken the assumption that pre-configured interrupts were valid, but doing a bit more digging this is false.  You're right, Linux does expect all IRQs to be disabled and masked, so this is a pinctrl-amd bug.

> So to me it seems that this holds for all interrupts marked as `WAKE_CNTRL_OFF_S0I3` by hardware, not just the SAM one. If I'm understanding all of this right, the proper solution would be to disable all wake pins at boot and then let `enable_irq_wake()` take care of it as (I believe) is expected.
> 
> 
> I think it'd be good if we had someone of the pinctrl/GPIO team look at this.

Agreed, I'll work on raising this with GPIO team.  I'm going to close this one.

PR #106, opened 2021-10-02
ACPI: PM: Include alternate AMDI0005 id in special behaviour

nakato:
The Surface Laptop 4 AMD has used the AMD0005 to identify this
controller instead of using the appropriate ACPI ID AMDI0005.  The
AMD0005 needs the same special casing as AMDI0005.

### DIFF
--- a/drivers/acpi/x86/s2idle.c
+++ b/drivers/acpi/x86/s2idle.c
@@ -371,7 +371,7 @@ static int lps0_device_attach(struct acpi_device *adev,
 		return 0;
 
 	if (acpi_s2idle_vendor_amd()) {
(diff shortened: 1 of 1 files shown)


nakato:
This builds on top of https://github.com/linux-surface/kernel/commit/8d39cf4e302f69eb34e21ce9a535fb3616fe45e6

I've posted the patches to the mailing list, so hopefully I can get them into a state that they can be included into mainline.
https://lore.kernel.org/linux-kernel/20211002041840.2058647-1-nakato@nakato.io/T/#t

qzed (approved):
Thanks!

PR #107, opened 2021-10-02
pinctrl: amd: disable and mask interrupts on probe

nakato:
Some systems such as the Microsoft Surface Laptop 4 leave interrupts
enabled and configured for use in sleep states on boot, which cause
unexpected behaviour such as spurious wakes and failed resumes in
s2idle states.

As interrupts should not be enabled until they are claimed and
explicitly enabled, disabling any interrupts mistakenly left enabled by
firmware should be safe.

I've proposed this upstream at: 
https://lore.kernel.org/linux-gpio/20211001161714.2053597-1-nakato@nakato.io/T/#t

### DIFF
--- a/drivers/pinctrl/pinctrl-amd.c
+++ b/drivers/pinctrl/pinctrl-amd.c
@@ -842,6 +842,32 @@ static const struct pinconf_ops amd_pinconf_ops = {
 	.pin_config_group_set = amd_pinconf_group_set,
 };
 
+static void amd_gpio_irq_init(struct amd_gpio *gpio_dev) {
+	struct pinctrl_desc *desc = gpio_dev->pctrl->desc;
+	unsigned long flags;
+	u32 pin_reg, mask;
+	int i;
+
+	mask = BIT(WAKE_CNTRL_OFF_S0I3) | BIT(WAKE_CNTRL_OFF_S3)
+		| BIT(INTERRUPT_MASK_OFF) | BIT(INTERRUPT_ENABLE_OFF)
+	        | BIT(INTERRUPT_MASK_OFF) | BIT(WAKE_CNTRL_OFF_S4);
+
+	for (i = 0; i < desc->npins; i++) {
+		int pin = desc->pins[i].number;
+		const struct pin_desc *pd = pin_desc_get(gpio_dev->pctrl, pin);
+		if (!pd)
+			continue;
+
+		raw_spin_lock_irqsave(&gpio_dev->lock, flags);
+
+		pin_reg = readl(gpio_dev->base + i * 4);
+		pin_reg &= ~mask;
+		writel(pin_reg, gpio_dev->base + i * 4);
+
+		raw_spin_unlock_irqrestore(&gpio_dev->lock, flags);
+	}
+}
+
 #ifdef CONFIG_PM_SLEEP
 static bool amd_gpio_should_save(struct amd_gpio *gpio_dev, unsigned int pin)
 {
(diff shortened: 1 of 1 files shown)


drivers/pinctrl/pinctrl-amd.c
```diff
@@ -842,6 +842,32 @@ static const struct pinconf_ops amd_pinconf_ops = {
...
+		| BIT(INTERRUPT_MASK_OFF) | BIT(INTERRUPT_ENABLE_OFF)
+	        | BIT(INTERRUPT_MASK_OFF) | BIT(WAKE_CNTRL_OFF_S4);
+
+	for (i = 0; i < desc->npins; i++) {
+		int pin = desc->pins[i].number;
+		const struct pin_desc *pd = pin_desc_get(gpio_dev->pctrl, pin);
+		if (!pd)
+			continue;
+
+		raw_spin_lock_irqsave(&gpio_dev->lock, flags);
```
qzed, line 861: Not sure if this locking is required or can maybe be avoided somehow since it's during probe. I think the pinctrl guys will tell you if that can be improved. For now it's better to be safe.

qzed (approved):
Looks good to me. Any reason to hold this back?

nakato:
I marked it as draft to take a conservative approach, as I'm not sure if it will get in upstream, but I'm happy for it to go in here.  It's made the ssh interrupts not be problematic anymore.

The main thing this patch fixes, in a Surface Laptop context, is spurious wakeups from the SSH interrupt, and a few other unknown interrupts that were enabled on boot but not enabled by Linux, and lacking any reference in the acpidump.  There is however an odd interaction between pinctrl-amd and the ACPI GPE's that means the device wakes up for things like un/plugging power, and closing the lid.  (Are laptops supposed to wake up when you close the lid?  It goes back to sleep after about 30 seconds which I think is systemd's doing.)

drivers/pinctrl/pinctrl-amd.c
nakato: They could probably be removed, if the registers were being modified at the same time as this operation that would probably mean something is configuring a pin, which would mean a possible race condition where a pin would be configured and then disabled here.

qzed: Right, good point with the race condition.

qzed:
> I marked it as draft to take a conservative approach, as I'm not sure if it will get in upstream, but I'm happy for it to go in here. It's made the ssh interrupts not be problematic anymore.

Sounds good to me. I'll merge this. We can always revert and apply the updated version later on once there are some comments on it.

> Are laptops supposed to wake up when you close the lid?

Partially, yes, but not completely. As far as I know, there's no way to differentiate if an interrupt is for close or open, so both should in some way wake the device. At some point there's supposed to be a check though that reads back the state of the lid. In the ACPI driver that's done in [`acpi_lid_update_state()`](https://elixir.bootlin.com/linux/v5.14.9/source/drivers/acpi/button.c#L358), which is called when the lid gets notified for a status change (e.g. [here](https://github.com/linux-surface/acpidumps/blob/f8db3d150815aa21530635b7e646eee271e3b8fe/surface_laptop_4_amd/ssdt9.dsl#L115), which triggers a call to [`acpi_button_notify()`](https://elixir.bootlin.com/linux/v5.14.9/source/drivers/acpi/button.c#L392) with `event == ACPI_BUTTON_NOTIFY_STATUS == 0x80`). Note the `if (state && signal_wakeup)` check in `acpi_lid_update_state()`, which should ensure that the wakeup event only gets sent when the lid is being opened.

So from what I can tell there are two things that might go wrong:
- The IRQ is sent before the lid state is being updated, i.e. IRQ for lid-close fires but state is still open when read in the IRQ handler.
- The IRQ triggers some other wakeup and it's not actually the lid driver that wakes the device up but maybe something else in the pinctrl/GPIO stack or some other driver listening to that IRQ.

I'm not an expert on this so though. Also not sure how the interaction between AC state and lid state should work, since that'd need to be done at higher level (like systemd maybe) as far as I can tell.

PR #113, opened 2021-11-04
Fixes for the ov8865 driver

djrscally:
Thanks to @jhautbois for spotting these. 

The black image problem people are having comes when the exposure set goes too close to the VTS. Most sensor drivers have some margin built in to prevent that, but I forgot to include that in the changes to the exposure control in ov8865. Add one in to avoid the black screen issue.

Opportunistically fix the max analogue gain. The field is 12 bits wide which is where the 8191 figure came from, but the driver specifies it as having a max of 16x gain


qzed (approved):
Thanks!

qzed:
I can confirm that the "black image problem" is fixed by this on my SB2.

PR #115, opened 2021-12-03
mfd: intel-lpss: Fix I2C4 not being available on the Microsoft Surface Go & Go 2

djrscally:
A patch from @jwrdegoede to fix the need to use the acpi_enforce_resources=lax boot param on the Go and Go2 - thanks Hans!


qzed (approved):
Thanks!

PR #116, opened 2022-01-07
platform/x86: int3472: Add board data for Surface Go 3

djrscally:
The Surface Go 3 needs some board data in order to configure the
TPS68470 PMIC - add entries to the tables in tps68470_board_data.c
that define the configuration that's needed.

Signed-off-by: Daniel Scally <djrscally@gmail.com>


qzed (approved):
Thanks!

PR #119, opened 2022-02-25
hid/multitouch: Add support for surface pro type cover tablet switch

PJungkamp:
The Surface Pro Type Cover has several non standard HID usages in it's
hid report descriptor.
I noticed that, upon folding the typecover back, a vendor specfic range
of 4 32 bit integer hid usages is transmitted.
Only the first byte of the message seems to convey reliable information
about the keyboard state.

0x22 => Normal (keys enabled)
0x33 => Folded back (keys disabled)
0x53 => Rotated left/right side up (keys disabled)
0x13 => Cover closed (keys disabled)
0x43 => Folded back and Tablet upside down (keys disabled)
This list may not be exhaustive.

The tablet mode switch will be disabled for a value of 0x22 and enabled
on any other value.

### DIFF
--- a/drivers/hid/hid-multitouch.c
+++ b/drivers/hid/hid-multitouch.c
@@ -76,6 +76,7 @@ MODULE_LICENSE("GPL");
 #define MT_QUIRK_FORCE_MULTI_INPUT	BIT(20)
 #define MT_QUIRK_DISABLE_WAKEUP		BIT(21)
 #define MT_QUIRK_HAS_TYPE_COVER_BACKLIGHT	BIT(22)
+#define MT_QUIRK_HAS_TYPE_COVER_TABLET_MODE_SWITCH	BIT(23)
 
 #define MT_INPUTMODE_TOUCHSCREEN	0x02
 #define MT_INPUTMODE_TOUCHPAD		0x03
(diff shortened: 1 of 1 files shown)


qzed:
Code looks good to me. Couple of style issues though and missing punctuation in comments (small things, really). Can you run checkpatch (`./scripts/checkpatch.pl --strict <patchfile>`, use `git format-patch -1` to get a patch) and address those remarks?

After that I think you could also try rebasing (needs a rebase due to the typecover backlight patch) and submitting it upstream.

PJungkamp:
Should I rebase against a revision before the backlight patch was applied?
I modified the code the backlight patch introduced since it fit the problem, so I depend on its presence. 
(I changed `get_type_cover_backlight_field(...)` to `get_type_cover_field(...)` since a field search is needed for both patches)

I'll correct the formatting within the next few days.

qzed:
> Should I rebase against a revision before the backlight patch was applied?

For upstreaming, yes. For the PR here I think it's better/easier if you keep it as-is.

PJungkamp:
I've rebased onto `5.17-surface-devel` and finally fixed the formatting issues.
I'll open a pull request for the `5.17-surface-devel` branch and close this one.

PR #122, opened 2022-04-28
Fix missing cameras in v5.17-surface*

djrscally:
This commit adds the missing VCM driver that's preventing the ipu3-cio2 driver from calling .complete() and building the media graph.

I should have moved the functionality from .complete() to .bound() so this would work regardless, but I'm tired and this was easier, so that annoying issue lives another day!


qzed (approved):
Thanks! Just tested and verified it's working.

PR #123, opened 2022-04-29
Add ancillary links

djrscally:
Ancillary links are a new kind of media link intended to describe the connection between media subdevs that are intended to operate as a single logical unit, principally flash LEDs and lens motors. Libcamera supports them already and includes auto focus algorithms now (at least on master) so with this series, the world facing camera should focus properly.


qzed:
Thanks, I'll give it a try in a second.

qzed (approved):

PR #124, opened 2022-05-06
media: ipu3-cio2: Move functionality from .complete() to .bound()

djrscally:
@qzed: Getting the VCM to defer probe is turning out to be quite annoying. In the interim, I finally got round to this. If the VCM defers then the focus won't work, but at least everything else still should.


qzed (approved):
Thanks!

PR #127, opened 2022-06-10
platform/x86/intel: hid: Add Surface Go to VGBS allow list

krnhotwings:
The Surface Go reports Chassis Type 9 (Laptop,) so the device needs to be
added to dmi_vgbs_allow_list to enable tablet mode when an attached Type
Cover is folded back.

Relevant bug report and discussion:
https://github.com/linux-surface/linux-surface/issues/837

Signed-off-by: Duke Lee <krnhotwings@gmail.com>


qzed (approved):

qzed:
Thanks!

PR #128, opened 2022-06-12
media: ipu3-cio2: Re-add .complete() to ipu3-cio2

djrscally:
This should fix the always-broken VCMs and allow cameras to work again everywhere. The VCM itself may still not work - it will depend on probe ordering, I still need to fix the missing propagation for the -EPROBE_DEFER call in the cio2-bridge. 

Sorry this took so long - I thought it was broken in some way as I couldn't get the focus to work but after extensive debugging it turned out I was just being an utter moron :D


qzed (approved):
No worries, thanks!

PR #130, opened 2022-11-10
add wdk2023 acpi to smmu platlist

nickdepinet:
We need to inform the SMMU driver of our ACPI for booting in ACPI mode to work on Windows Dev Kit 2023

### DIFF
--- a/drivers/iommu/arm/arm-smmu/arm-smmu-qcom.c
+++ b/drivers/iommu/arm/arm-smmu/arm-smmu-qcom.c
@@ -446,6 +446,7 @@ static const struct of_device_id __maybe_unused qcom_smmu_impl_of_match[] = {
 static struct acpi_platform_list qcom_acpi_platlist[] = {
 	{ "LENOVO", "CB-01   ", 0x8180, ACPI_SIG_IORT, equal, "QCOM SMMU" },
 	{ "QCOM  ", "QCOMEDK2", 0x8180, ACPI_SIG_IORT, equal, "QCOM SMMU" },
+	{ "QCOM  ", "QCOMEDK2", 0x8280, ACPI_SIG_IORT, equal, "QCOM SMMU" },
 	{ }
 };
 #endif


drivers/iommu/arm/arm-smmu/arm-smmu-qcom.c
```diff
@@ -446,6 +446,7 @@ static const struct of_device_id __maybe_unused qcom_smmu_impl_of_match[] = {
 static struct acpi_platform_list qcom_acpi_platlist[] = {
 	{ "LENOVO", "CB-01   ", 0x8180, ACPI_SIG_IORT, equal, "QCOM SMMU" },
 	{ "QCOM  ", "QCOMEDK2", 0x8180, ACPI_SIG_IORT, equal, "QCOM SMMU" },
+	{ "QCOM  ", "QCOMEDK2", 0x8280, ACPI_SIG_IORT, equal, "QCOM SMMU" },
```
qzed, line 449: Are you sure that `QCOM`, `QCOMEDK2` is the right combination here? AFAIK that's for Qualcomm's developer kit. Although it could be possible that MS is re-using that.

nickdepinet: Here's a photo of the ACPI logs from boot attempt: If you've got instructions on how to get a dump out of windows /wsl I can do that too 
acpi_dump

qzed: `acpidump` and other ACPICA tools are actually also available for windows: https://acpica.org/downloads/binary-tools. So you can use that directly from Windows. However, before posting don't forget to filter out the MSDM section containing some Windows licensing information (e.g. as described here: https://github.com/linux-surface/linux-surface/wiki/Testers-and-Contributing#contributing-acpidumps.

qzed:
Thanks!

PR #132, opened 2022-12-11
HID based IPTS and ITHC support

StollD:
iptsd feels pretty much ready and will work better than what is currently in the repo, so lets start doing this ...

Things that need to be done once this is merged:
* Update `CONFIG_MISC_IPTS=m` to `CONFIG_HID_IPTS=m` in our kernel configs
* Add `CONFIG_HID_ITHC=m` to the kernel configs
* Bump iptsd version to 0.6 and tag a release

Why do this now if 6.1 is right around the corner? Because then we would have to wait until Fedora rebases to 6.1, or apply it to 6.0 anyways, because the new iptsd is incompatible with the old driver.


StollD:
@quo I know there were some suspend problems on ITHC, do you think its ready? Or should we wait with this?

drivers/hid/ipts/cmd.c
```diff
@@ -0,0 +1,144 @@
...
+	if (ret <= 0) {
+		dev_err(ipts->dev, "Error while reading response: %d\n", ret);
+		return ret;
+	}
+
+	/*
+	 * In a response, the command code will have the most significant bit
+	 * flipped to one. We check for this and then set the bit to 0.
+	 */
+	if ((rsp.cmd & 0x80000000) == 0) {
```
qzed, line 69: Maybe consider adding a #define for that "magic value"?

drivers/hid/ithc/ithc-regs.h
```diff
@@ -0,0 +1,186 @@
...
+		/* 1130/1230 */ u64 _unknown_1130_guc_addr;
+		/* 1138/1238 */ u32 _unknown_1138_guc;
+		/* 113c/123c */ u32 _unknown_113c;
+		/* 1140/1240 */ u32 _unknown_1140_guc;
+		/* 1144/1244 */ u32 _unknown_1144[23];
+		/* 11a0/12a0 */ u32 _unknown_11a0_counters[6];
+		/* 11b8/12b8 */ u32 _unknown_11b8[18];
+	} dma_rx[2];
+};
+_Static_assert(sizeof(struct ithc_registers) == 0x1300);
```
qzed, line 131: Maybe mark this as `__packed`? Also I think plain `static_assert` is preferred in the kernel.

qzed (commented):
Nice work! I have some small remarks and one potential concurrency issue, but I don't think that's anything that should prevent a merge.

drivers/hid/ipts/spec-device.h
```diff
@@ -0,0 +1,279 @@
...
+ * @max_contacts:  Maximum amount of concurrent touches the sensor can process.
+ */
+struct ipts_device_info {
+	u16 vendor;
+	u16 product;
+	u32 hw_version;
+	u32 fw_version;
+	u32 data_size;
+	u32 feedback_size;
+	enum ipts_mode mode;
```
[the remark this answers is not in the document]
StollD: It used to be that way but it was hell to document since now you have all these implicit references to the enums sprinkled through the structs. Using the enum directly avoids that, and makes the relation between the value and where it should go more clear. Thats kinda important to me, because that header file is pretty much the only docs on IPTS that we have. 

To prevent any issues, I added the static_asserts below the structs which would make the build fail if the compiler decided to use a different size for the enums, and so far it didnt cause any issues.

drivers/hid/ipts/cmd.c
StollD: To be honest, I dont think we need a define for a simple bitmask, but changing it to `BIT(31)` would be a good idea.

qzed: I mean you already have the `IPTS_CMD_...` stuff, why not add a `#define IPTS_CMD_RESPONSE_BIT BIT(31)` as well since you already use it twice? Reads a bit nicer IMHO and ties in a bit better to `spec-device.h` being a more complete representation of the spec.

drivers/hid/ipts/spec-device.h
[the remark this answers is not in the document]
qzed: Alright. I mean you could add a comment behind the field pointing to the enum. I think the C spec is rather undefined/non-explicit on that unfortunately, but you're right, the `static_assert()` should normally catch any problems. I mostly mentioned this because I have a bit of a mistrust towards compilers sometimes.

drivers/hid/ithc/ithc-api.h
```diff
@@ -0,0 +1,16 @@
...
+	struct miscdevice m;
+	atomic_t open_count;
+};
+
+struct ithc_api_header {
+	u8 hdr_size;
+	u8 reserved[3];
+	u32 msg_num;
+	u32 size;
+};
```
qzed, line 13: Should this also be `__packed`?

quo:
> I know there were some suspend problems on ITHC recently, do you think its ready? Or should we wait with this?

It's ready as far as I'm concerned.

To summarize the suspend issue: Before I added the dev_pm_ops struct (for hibernation), I don't think Linux was suspending the THC at all. Now it is suspending it, causing the touch controller to lose state. I wasn't able to figure out a minimal way to restore state (or to avoid losing state) that works on all devices, so now the driver basically just shuts down on suspend and completely restarts on wake. This is effectively what the Windows driver does as well. The only minor downside is that the HID device is recreated, so iptsd has to restart, but that seems to work fine now.

drivers/hid/ithc/ithc-regs.h
quo: Yeah, I'll change it to `static_assert`. I prefer to use `packed` only for stuff that's actually misaligned.

drivers/hid/ithc/ithc-api.h
quo: This is the old non-HID API. I've kept it around for debugging, but I'll remove all of it in the near future.

qzed:
I believe this is technically ready? The last issues are just style things (and if we're going to upstream this that's probably something that we can work on then / when it's merged). So I'd merge it this weekend if there are no other complaints.

jchv:
Is it me, or does `ithc-main.c` contain carriage returns? I'm seeing `^M`s all over in `git diff`. I assume this is unintentional :)

qzed:
It's not just you. I didn't notice that because GitHub doesn't show it... I can fix that up after merging if I don't forget. Anyway, the ITHC code does need quite a bit of work/formatting before we can consider it for upstream.

jchv:
Cool. I'm testing it now on my Tigerlake Surface Laptop 4 and it works fine so far, including suspend and resume. (Just single touch though; I haven't figured out how to get iptsd@master to work with the ithc driver from this patch, if it's currently possible.)

StollD:
> I can fix that up after merging if I don't forget.

I'd recommend doing any changes or fixups on quos repository, so that we don't diverge and updating is as simple as `cp -r src/* drivers/hid/ithc/` 

> edit: Figured it out. For some reason, I needed to specify the `hid` parameter explicitly for it to use HID mode... Not sure why. It seems to me like it should be loading in HID mode by default based on this patch.

If you can still replicate it, I would be interested in your dmesg when loading ithc without the parameter.

qzed:
In that case, should we also move that repo to linux-surface/ithc?

quo:
> Is it me, or does `ithc-main.c` contain carriage returns?

Yeah, I've been meaning to fix that.

> In that case, should we also move that repo to linux-surface/ithc?

Our company is migrating away from the Surface devices, and we may not even end up using the SP7+ in production. So my plan is to clean up the ithc sources a bit (maybe over the holidays), and after that I'm happy to transfer the repository and hand over maintenance to someone else.

qzed:
Just to be clear what I meant: I'd add you to this org, then you can transfer it and still work on it here (no strings attached). It's just to make this a bit more "official". However, it's fully up to you.

quo:
That's very kind. I'll let you know if I change my mind, but I just don't see myself spending a lot more time on the driver (esp. upstreaming, etc), so I think it would be better to just hand it over eventually.

qzed:
Sure, whatever you're more comfortable with.

StollD:
Marking this as a draft until I have a proper fix for a suspend issue on SP5 that verdre mentioned on Matrix.

StollD:
Superceded by #133

PR #133, opened 2023-01-21
HID based IPTS and ITHC support, 6.1 edition

StollD:
iptsd feels pretty much ready and will work better than what is currently in the repo, so lets start doing this ...

Things that need to be done once this is merged:
* Update `CONFIG_MISC_IPTS=m` to `CONFIG_HID_IPTS=m` in our kernel configs
* Add `CONFIG_HID_ITHC=m` to the kernel configs
* Bump iptsd version to 0.6 and tag a release

Changes since #132:
 * ITHC: https://github.com/quo/ithc-linux/compare/67ebf094f271478dd4ba3dddc210ca779bac9f31...e1c3978daefee9a944270c3b059534d5227af3e8
 * IPTS: https://github.com/linux-surface/intel-precise-touch/compare/f4c781af0908fbf24517de97235e788283f363af...19e8512a8110186fe772fbdc534c98139d65943d


drivers/hid/ipts/thread.h
```diff
@@ -0,0 +1,62 @@
...
+ * to issue MEI commands from that thread while it shuts itself down. By using a custom
+ * boolean variable and a completion object, we can call kthread_stop only when the thread
+ * already finished all of its work and has returned.
+*/
+struct ipts_thread {
+	struct task_struct *thread;
+
+	bool should_stop;
+
+	struct mutex lock;
```
qzed, line 27: You could replace those with set_bit() and test_bit() (https://www.kernel.org/doc/html/latest/core-api/kernel-api.html?highlight=set_bit#bit-operations) as full mutexes are a bit overkill for that task. Alternatives are atomics (which is probably better for explicit acquire/release semantics without needing rmb/wmb) or I guess you could even use `READ_ONCE()` and `WRITE_ONCE()` with memory barriers if you want to keep the bool.

So with bitops you'd do something like:
```c
int ipts_thread_stop(struct ipts_thread *thread)
{
	int ret = 0;

	if (!thread)
		return -EFAULT;

	if (!thread->thread)
		return 0;

	set_bit(IPTS_THREAD_STOP_BIT, &thread->flags);
	wmb();  // ensure bit is set before we start waiting

	wait_for_completion(&thread->done);
	ret = kthread_stop(thread->thread);

	thread->thread = NULL;
	thread->data = NULL;
	thread->threadfn = NULL;

	return ret;
}
```
and
```c
bool ipts_thread_should_stop(struct ipts_thread *thread)
{
	bool ret = false;

	if (!thread)
		return false;

	// optional: add rmb(); here to have "acquire" semantics, but that's not really necessary
	return test_bit(IPTS_THREAD_STOP_BIT, thread->flags);
}
```

qzed (approved):
Looks good to me (I haven't extensively checked it though). I've added one comment for a potential optimization, but that's something that we can work on later.

drivers/hid/ipts/thread.h
StollD: Would `smp_load_acquire` / `smp_store_release` work too? That could be the easiest solution, because it already seems to combine `READ_ONCE` and `WRITE_ONCE` with a barrier.

qzed: Oh, I forgot about those. Yeah I think those should work too.

qzed: Although... After checking, those have the barriers exactly the opposite way (which is what you more often want), but I think we only need the `wmb()` after setting the flag. If at all. I'm not entirely sure there how the semantics are for the `wait_for_completion()` call, but the point is that we want to make sure that the bit is set before waiting.

StollD: Done

PR #135, opened 2023-03-02
platform/x86: int3472: Add GPIOs to Surface Go 3 Board data

djrscally:
Add the INT347E GPIO lookup table to the board data for the Surface Go 3. This is necessary to allow the ov7251 IR camera to probe properly on that platform.

Fixes https://github.com/linux-surface/linux-surface/issues/1055


qzed (approved):
Thanks!

PR #137, opened 2023-03-22
Add support for camera LEDs on Surface Go platforms

djrscally:
This series adds support for the privacy LEDs on the Surface Go line; the world facing camera's white indicator LED should work (all thanks to Kate and Hans), as should the IR LED for the IR camera.


qzed:
Thanks!

PR #138, opened 2023-03-23
Revert LED support for the Go2

djrscally:
Let's take this out for now in the end. On windows the LED is re-triggered constantly in flash mode, so I followed that behaviour, but someones on the mailing list is concerned at the potential for hardware damage doing that...given on Windows it's only going to be active for as long as it takes to log in through Windows Hello but there's no such restriction in Linux it's a good point.


qzed:
Oh right, that make sense. Thanks!

PR #139, opened 2023-04-28
Update IPTS from module repo

StollD:
Changes:
 * Fix redefinition error on AOSP clang
 * Increase the polling frequency to reduce latency
 * Don't allocate a new buffer for every HID report
 * Always use the generic HID driver instead of forcing hid-multitouch

Based on https://github.com/linux-surface/intel-precise-touch/commit/a2b675d72dbde80ebe36a5b6ceaebd596c030314


qzed (approved):
Thanks!

PR #140, opened 2023-05-11
Correct fix for memory pagefault when using cameras

djrscally:
Hey - here's the proper fix for the camera pagefault. We probably also want to fix the IPU3 driver since it never really should have used v4l2_subdev_link_validate(), but that's a separate thing.


qzed (approved):
Thanks!

PR #142, opened 2023-07-19
Update IPTS from module repo

StollD:
Changes:
* Remove usages of C11 for compatibility with older (Android) kernels
* Remove sensor reset during shutdown
* Fix compiling resources.c on Android-x86 kernel
* Split out HID behaviour for different EDS versions into different files
* Fix kernel-doc comments
* Add missing docs for members of the device info struct
* Drop Intel copyright
* Rename doorbell mode to poll mode
* Disable the HID interface when the hardware is shutting down.

Based on https://github.com/linux-surface/intel-precise-touch/commit/56765da3f87d87c5e5cf09e928abe0f579525e5e


qzed (approved):
Thanks!

PR #145, opened 2023-12-17
Platform profile: switch fan profile

iwanders:
Followup from https://github.com/linux-surface/kernel/pull/144 that adds support to switch the fan profile whenever the platform profile is switched.

Values recorded with IRPMon as sent to the FAN together with the numbers sent to the TMP module:
```
2023-12-16 6:54:41 PM   Seq OUT TMP:03 t01 i00 c 0x03 (  4): 01 00 00 00 
2023-12-16 6:54:41 PM   Seq OUT FAN:05 t01 i01 c 0x0e (  1): 02 

2023-12-16 6:54:45 PM   Seq OUT TMP:03 t01 i00 c 0x03 (  4): 02 00 00 00 
2023-12-16 6:54:45 PM   Seq OUT FAN:05 t01 i01 c 0x0e (  1): 01 

2023-12-16 6:54:30 PM   Seq OUT TMP:03 t01 i00 c 0x03 (  4): 03 00 00 00 
2023-12-16 6:54:30 PM   Seq OUT FAN:05 t01 i01 c 0x0e (  1): 03 

2023-12-16 6:54:33 PM   Seq OUT TMP:03 t01 i00 c 0x03 (  4): 04 00 00 00 
2023-12-16 6:54:33 PM   Seq OUT FAN:05 t01 i01 c 0x0e (  1): 04 
```

I've confirmed the changes in this PR using the following methodology;
- Switch platform profile in gnome (and with `echo balanced-performance > /sys/firmware/acpi/platform_profile` for normal)
- Start recording fan speed.
- Start `stress -c 12`.
- Cancel `stress -c 12` after steady state has been reached, plus a bit (didn't time this). Mark time.
- Wait for fan to become completely idle again.

This obtains us the following fan curves:

comparison

We clearly get 4 different fan max speeds, with performance as the highest level.

### DIFF
--- a/drivers/platform/surface/surface_aggregator_registry.c
+++ b/drivers/platform/surface/surface_aggregator_registry.c
@@ -68,10 +68,24 @@ static const struct software_node ssam_node_bat_sb3base = {
 	.parent = &ssam_node_hub_base,
 };
 
-/* Platform profile / performance-mode device. */
-static const struct software_node ssam_node_tmp_pprof = {
+/* Platform profile / performance-mode device without a fan. */
+static const struct software_node ssam_node_tmp_perf_profile = {
 	.name = "ssam:01:03:01:00:01",
 	.parent = &ssam_node_root,
+};
+
+/* Platform profile / performance-mode device with a fan, such that
+ * the fan controller profile can also be switched.
+ */
+static const struct property_entry ssam_node_tmp_perf_profile_has_fan[] = {
+	PROPERTY_ENTRY_BOOL("has_fan"),
+	{ }
+};
+
+static const struct software_node ssam_node_tmp_perf_profile_with_fan = {
+	.name = "ssam:01:03:01:00:01",
+	.parent = &ssam_node_root,
+	.properties = ssam_node_tmp_perf_profile_has_fan,
 };
 
 /* Tablet-mode switch via KIP subsystem. */
@@ -202,7 +216,7 @@ static const struct software_node ssam_node_pos_tablet_switch = {
  */
 static const struct software_node *ssam_node_group_gen5[] = {
 	&ssam_node_root,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	NULL,
 };
 
@@ -213,7 +227,7 @@ static const struct software_node *ssam_node_group_sb3[] = {
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
 	&ssam_node_bat_sb3base,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	&ssam_node_bas_dtx,
 	&ssam_node_hid_base_keyboard,
 	&ssam_node_hid_base_touchpad,
@@ -227,7 +241,7 @@ static const struct software_node *ssam_node_group_sl3[] = {
 	&ssam_node_root,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	&ssam_node_hid_main_keyboard,
 	&ssam_node_hid_main_touchpad,
 	&ssam_node_hid_main_iid5,
@@ -239,7 +253,7 @@ static const struct software_node *ssam_node_group_sl5[] = {
 	&ssam_node_root,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	&ssam_node_hid_main_keyboard,
 	&ssam_node_hid_main_touchpad,
 	&ssam_node_hid_main_iid5,
@@ -252,7 +266,7 @@ static const struct software_node *ssam_node_group_sls1[] = {
 	&ssam_node_root,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	&ssam_node_pos_tablet_switch,
 	&ssam_node_hid_sam_keyboard,
 	&ssam_node_hid_sam_penstash,
@@ -268,7 +282,7 @@ static const struct software_node *ssam_node_group_sls2[] = {
 	&ssam_node_root,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	&ssam_node_pos_tablet_switch,
 	&ssam_node_hid_sam_keyboard,
 	&ssam_node_hid_sam_penstash,
@@ -282,7 +296,7 @@ static const struct software_node *ssam_node_group_slg1[] = {
 	&ssam_node_root,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	NULL,
 };
 
@@ -291,7 +305,7 @@ static const struct software_node *ssam_node_group_sp7[] = {
 	&ssam_node_root,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	NULL,
 };
 
@@ -301,7 +315,7 @@ static const struct software_node *ssam_node_group_sp8[] = {
 	&ssam_node_hub_kip,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile,
 	&ssam_node_kip_tablet_switch,
 	&ssam_node_hid_kip_keyboard,
 	&ssam_node_hid_kip_penstash,
@@ -318,7 +332,7 @@ static const struct software_node *ssam_node_group_sp9[] = {
 	&ssam_node_hub_kip,
 	&ssam_node_bat_ac,
 	&ssam_node_bat_main,
-	&ssam_node_tmp_pprof,
+	&ssam_node_tmp_perf_profile_with_fan,
 	&ssam_node_pos_tablet_switch,
 	&ssam_node_hid_kip_keyboard,
 	&ssam_node_hid_kip_penstash,
--- a/drivers/platform/surface/surface_platform_profile.c
+++ b/drivers/platform/surface/surface_platform_profile.c
@@ -1,7 +1,7 @@
 // SPDX-License-Identifier: GPL-2.0+
 /*
  * Surface Platform Profile / Performance Mode driver for Surface System
- * Aggregator Module (thermal subsystem).
+ * Aggregator Module (thermal and fan subsystem).
  *
  * Copyright (C) 2021-2022 Maximilian Luz <luzmaximilian@gmail.com>
  */
@@ -14,11 +14,22 @@
 
 #include <linux/surface_aggregator/device.h>
 
+// Enum for the platform performance profile sent to the TMP module.
 enum ssam_tmp_profile {
 	SSAM_TMP_PROFILE_NORMAL             = 1,
 	SSAM_TMP_PROFILE_BATTERY_SAVER      = 2,
 	SSAM_TMP_PROFILE_BETTER_PERFORMANCE = 3,
 	SSAM_TMP_PROFILE_BEST_PERFORMANCE   = 4,
+};
+
+// Enum for the fan profile sent to the FAN module. This fan profile is
+// only sent to the EC if the 'has_fan' property is set. The integers are
+// not a typo, they differ from the performance profile indices.
+enum ssam_fan_profile {
+	SSAM_FAN_PROFILE_NORMAL             = 2,
+	SSAM_FAN_PROFILE_BATTERY_SAVER      = 1,
+	SSAM_FAN_PROFILE_BETTER_PERFORMANCE = 3,
+	SSAM_FAN_PROFILE_BEST_PERFORMANCE   = 4,
 };
 
 struct ssam_tmp_profile_info {
@@ -27,9 +38,10 @@ struct ssam_tmp_profile_info {
 	__le16 unknown2;
 } __packed;
 
-struct ssam_tmp_profile_device {
+struct ssam_platform_profile_device {
 	struct ssam_device *sdev;
 	struct platform_profile_handler handler;
+	bool has_fan;
 };
 
 SSAM_DEFINE_SYNC_REQUEST_CL_R(__ssam_tmp_profile_get, struct ssam_tmp_profile_info, {
@@ -40,6 +52,13 @@ SSAM_DEFINE_SYNC_REQUEST_CL_R(__ssam_tmp_profile_get, struct ssam_tmp_profile_in
 SSAM_DEFINE_SYNC_REQUEST_CL_W(__ssam_tmp_profile_set, __le32, {
 	.target_category = SSAM_SSH_TC_TMP,
 	.command_id      = 0x03,
+});
+
+SSAM_DEFINE_SYNC_REQUEST_W(__ssam_fan_profile_set, char, {
+	.target_category = SSAM_SSH_TC_FAN,
+	.target_id = SSAM_SSH_TID_SAM,
+	.command_id = 0x0e,
+	.instance_id = 0x01,
 });
 
 static int ssam_tmp_profile_get(struct ssam_device *sdev, enum ssam_tmp_profile *p)
@@ -62,7 +81,14 @@ static int ssam_tmp_profile_set(struct ssam_device *sdev, enum ssam_tmp_profile 
 	return ssam_retry(__ssam_tmp_profile_set, sdev, &profile_le);
 }
 
-static int convert_ssam_to_profile(struct ssam_device *sdev, enum ssam_tmp_profile p)
+static int ssam_fan_profile_set(struct ssam_device *sdev, enum ssam_fan_profile p)
+{
+	char profile = p;
+
+	return ssam_retry(__ssam_fan_profile_set, sdev->ctrl, &profile);
+}
+
+static int convert_ssam_tmp_to_profile(struct ssam_device *sdev, enum ssam_tmp_profile p)
 {
 	switch (p) {
 	case SSAM_TMP_PROFILE_NORMAL:
@@ -83,7 +109,8 @@ static int convert_ssam_to_profile(struct ssam_device *sdev, enum ssam_tmp_profi
 	}
 }
 
-static int convert_profile_to_ssam(struct ssam_device *sdev, enum platform_profile_option p)
+
+static int convert_profile_to_ssam_tmp(struct ssam_device *sdev, enum platform_profile_option p)
 {
 	switch (p) {
 	case PLATFORM_PROFILE_LOW_POWER:
@@ -105,20 +132,42 @@ static int convert_profile_to_ssam(struct ssam_device *sdev, enum platform_profi
 	}
 }
 
+static int convert_profile_to_ssam_fan(struct ssam_device *sdev, enum platform_profile_option p)
+{
+	switch (p) {
+	case PLATFORM_PROFILE_LOW_POWER:
+		return SSAM_FAN_PROFILE_BATTERY_SAVER;
+
+	case PLATFORM_PROFILE_BALANCED:
+		return SSAM_FAN_PROFILE_NORMAL;
+
+	case PLATFORM_PROFILE_BALANCED_PERFORMANCE:
+		return SSAM_FAN_PROFILE_BETTER_PERFORMANCE;
+
+	case PLATFORM_PROFILE_PERFORMANCE:
+		return SSAM_FAN_PROFILE_BEST_PERFORMANCE;
+
+	default:
+		/* This should have already been caught by platform_profile_store(). */
+		WARN(true, "unsupported platform profile");
+		return -EOPNOTSUPP;
+	}
+}
+
 static int ssam_platform_profile_get(struct platform_profile_handler *pprof,
 				     enum platform_profile_option *profile)
 {
-	struct ssam_tmp_profile_device *tpd;
+	struct ssam_platform_profile_device *tpd;
 	enum ssam_tmp_profile tp;
 	int status;
 
-	tpd = container_of(pprof, struct ssam_tmp_profile_device, handler);
+	tpd = container_of(pprof, struct ssam_platform_profile_device, handler);
 
 	status = ssam_tmp_profile_get(tpd->sdev, &tp);
 	if (status)
 		return status;
 
-	status = convert_ssam_to_profile(tpd->sdev, tp);
+	status = convert_ssam_tmp_to_profile(tpd->sdev, tp);
 	if (status < 0)
 		return status;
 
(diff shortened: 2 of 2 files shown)


drivers/platform/surface/surface_platform_profile.c
```diff
@@ -30,6 +39,7 @@ struct ssam_tmp_profile_info {
 struct ssam_tmp_profile_device {
 	struct ssam_device *sdev;
 	struct platform_profile_handler handler;
+	bool has_fan;
```
iwanders, line 42: Technically, we could call `device_property_read_bool` each time the profile is changed, but storing it as a local setting felt reasonable.

drivers/platform/surface/surface_platform_profile.c
```diff
@@ -42,6 +52,13 @@ SSAM_DEFINE_SYNC_REQUEST_CL_W(__ssam_tmp_profile_set, __le32, {
 	.command_id      = 0x03,
 });
 
+SSAM_DEFINE_SYNC_REQUEST_W(__ssam_fan_profile_set, char, {
+	.target_category = SSAM_SSH_TC_FAN,
+	.target_id = SSAM_SSH_TID_SAM,
+	.command_id = 0x0e,
+	.instance_id = 0x01,
+});
```
iwanders, line 60: I couldn't use the `SSAM_DEFINE_SYNC_REQUEST_CL_R` here, as that would send commands to the instance id of the `TMP` module, this works, but I'm not sure if it's the best way to solve that.

drivers/platform/surface/surface_platform_profile.c
```diff
@@ -83,7 +107,8 @@ static int convert_ssam_to_profile(struct ssam_device *sdev, enum ssam_tmp_profi
 	}
 }
 
-static int convert_profile_to_ssam(struct ssam_device *sdev, enum platform_profile_option p)
+
+static int convert_profile_to_ssam_tmp(struct ssam_device *sdev, enum platform_profile_option p)
```
iwanders, line 111: Decided to rename this and a few other things to ensure it is clear it pertains to the `tmp` profile not the `fan` profile.

drivers/platform/surface/surface_platform_profile.c
```diff
@@ -105,6 +130,28 @@ static int convert_profile_to_ssam(struct ssam_device *sdev, enum platform_profi
 	}
 }
 
+static int convert_profile_to_ssam_fan(struct ssam_device *sdev, enum platform_profile_option p)
```
iwanders, line 133: An alternative to this function would be to just use the `ssam_tmp_profile` to index into a lookup table to convert to the fan profile. But I went this route in case we end up wanting to expose decoupled fan profile / platform profile support.

iwanders (commented):
@qzed , filed a draft PR for the additions to the platform profile, still some outstanding questions here and there on the approach, but the overall functionality is there.

drivers/platform/surface/surface_platform_profile.c
qzed: Yeah, I think it's generally better to read properties once and store them.

drivers/platform/surface/surface_platform_profile.c
qzed: Hmm, that is actually a good question. Windows uses IID 1, correct?

This also kind of begs the question how multiple fans would be handled here. I.e., whether each would have its own fan profile or whether there is a single fan profile for all fans... If there would be only a single profile, IID 0 should be preferred, otherwise the IID of the respective fan.

Unfortunately, that's not something that we can easily answer I think (if at all). So I guess for now go with whatever Windows uses.

drivers/platform/surface/surface_platform_profile.c
qzed: Hmm, `tmp` is kind of the name of the whole subsystem. Maybe `perf` is better (i.e. performance vs. fan profile)? To me it's still somewhat of a mystery what this thing actually does on the different platforms. If you can figure that out (or something like what it's called on Windows), we could use the name of that.

drivers/platform/surface/surface_platform_profile.c
iwanders: Yeah, windows uses instance id 1, from the description;
```
2023-12-16 6:54:41 PM   Seq OUT FAN:05 t01 i01 c 0x0e (  1): 02 
```
`i01` [stands for](https://github.com/iwanders/nixos-surface/blob/a3196593eb51deef04f6e5152f5871b048ace30a/dev/irp_display.py#L274-L286) instance id 01.

I guess we'll find out how it works with multiple fans if they ever make a surface with two fans? :D  Is the surface pro 9 the first one with a fan?, or have others had fans but no one really dug into handling them?

Should I explore whether sending to instance id 0 works? Is that the 'apply all all instances' flag?

drivers/platform/surface/surface_platform_profile.c
iwanders: Yeah, perf is better, agreed, I'll change that over.

> To me it's still somewhat of a mystery what this thing actually does on the different platforms. If you can figure that out (or something like what it's called on Windows), we could use the name of that.

I spent the past hour or so spelunking in various `.sys` files... I don't think the fan profiles (or the whole `tmp` category) are named anywhere, I expect they're an enum in the original headers and just evaporate into a few integers somewhere in the drivers. I also can't find the limits we see in the plots in the firmware.

Interesting bits from looking for strings though;
```
SurfaceUpdate/surfacesystemmanagementframework$ strings -n 5 SurfaceSystemManagementFrameworkDriver.sys  | grep -i fan
SmfClientControl_FanControlCoreConfigurationGet{Use fan control channels with configuration}[Manual]
SmfClientControl_FanControlCoreConfigurationGet{Use fan control channels with configuration}
```
Looks like that file contains some test routines, another one is:
```
SmfProcessing_TimerPrepareLocked.MaximumProcessingTimeUpdate{Heat up the unit so that PID starts. This may take 20 minutes. Then cool the unit down so that timer stops. After that heat it again.}[Manual]
```
I wonder if there's a way to trigger that 'action' or something from Windows.


I also tried to set the fan profile and then retrieve various registers. I actually expected CID 6 ([coefficients?](https://github.com/iwanders/nixos-surface/blob/a3196593eb51deef04f6e5152f5871b048ace30a/fan.md#cid-6-coefficients)) to change, but it doesn't, what I did notice is that CID 13 does give use a float back that changes:

```
$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 0x0e 1 0 1
$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 13 1 1 
00 00 c8 41
Little endian f32: 25.0

$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 0x0e 1 0 2
$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 13 1 1 
00 00 f0 41
Little endian f32: 30.0

$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 0x0e 1 0 3
$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 13 1 1 
00 00 00 42
Little endian f32: 32.0

$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 0x0e 1 0 4
$ /home/ivor/.nix-profile/bin/python ./ctrl.py request 5 1 13 1 1 
00 00 10 42
Little endian f32: 36.0
```

I wouldn't be surprised if this is the controllers' target temperature for one of the sensors. So _technically_ there's a way to query the fan profile, but I'm not sure how much value that provides.

Anyways, I'll go make the `tmp` to `perf` change after lunch.

iwanders: Okay, not sure if this is actually what you had in mind in hindsight, but  d385b72fc03d8929b529a83e9ad40e6fe4d99c14  and 54d09a8c1  changes from `tmp` to `perf`. It felt weird to have a `convert_profile_to_ssam_perf` function that returned a `SSAM_TMP_PROFILE_BATTERY_SAVER` enum version. So instead I replaced `TMP` with `PERF` there as well...

It also renames the device struct from `ssam_tmp_profile_device` to `ssam_platform_profile_device`, since it handles more than `tmp` now, and actually handles the platform profile, with both changes to the TMP subsystem as well as the FAN subystem?

This does make the PR significantly larger, but perhaps it clears up the functionality / naming. Thoughts very welcome, I'm happy to revert if this is too extensive, I personally was hoping for a relatively small PR here, since it feels like we're bolting on just a bit of functionality. But perhaps this is the right way, your call.

qzed: Yeah that makes sense and I think overall will result in a cleaner/more understandable module.

iwanders: Okay, based on this I've also renamed the device node in the registry from `ssam_node_tmp_pprof` to `ssam_node_perf_profile`. The original `ssam_node_tmp_pprof_with_fan` just sounded wrong.

iwanders:
I've marked this as ready for review. I cleaned up some comments here and there, and renamed the node in the registry. I've also squashed the commits and put a reasonable commit message / title on it.

drivers/platform/surface/surface_platform_profile.c
qzed: > I guess we'll find out how it works with multiple fans if they ever make a surface with two fans? :D Is the surface pro 9 the first one with a fan?, or have others had fans but no one really dug into handling them?

I suspect that the SLS and some other of the new devices also have fan control support (an easy way to check seems to be the ACPI PNP-whatever FAN node). The SB2 has a fan in the base for the dGPU, but all I found for controlling that was the performance profile. So on the newer devices I think just no one looked into this before you.

> Should I explore whether sending to instance id 0 works? Is that the 'apply all all instances' flag?

I think it's probably best to stick with whatever windows does for now.

drivers/platform/surface/surface_platform_profile.c
```diff
@@ -83,20 +109,43 @@ static int convert_ssam_to_profile(struct ssam_device *sdev, enum ssam_tmp_profi
...
+	case PLATFORM_PROFILE_BALANCED_PERFORMANCE:
+		return SSAM_PERF_PROFILE_BETTER_PERFORMANCE;
+
+	case PLATFORM_PROFILE_PERFORMANCE:
+		return SSAM_PERF_PROFILE_BEST_PERFORMANCE;
+
+	default:
+		/* This should have already been caught by platform_profile_store(). */
+		WARN(true, "unsupported platform profile");
+		return -EOPNOTSUPP;
```
qzed, line 131: I'm not sure if `EOPNOTSUPP` is the best value here. Maybe `EINVAL` fits better?

drivers/platform/surface/surface_aggregator_registry.c
```diff
@@ -68,12 +68,26 @@ static const struct software_node ssam_node_bat_sb3base = {
 	.parent = &ssam_node_hub_base,
 };
 
-/* Platform profile / performance-mode device. */
-static const struct software_node ssam_node_tmp_pprof = {
+/* Platform profile / performance-mode device without a fan. */
+static const struct software_node ssam_node_perf_profile = {
```
qzed, line 72: So I thought about this a bit and I'd prefer to keep the `tmp` prefix as indicator for the subsystem in the registry. So `ssam_node_SUBSYS_NAME`, e.g., `ssam_node_tmp_perf_profile`. I'll add a comment for this to the fan monitor patches you submitted to the mailing list as well.

drivers/platform/surface/surface_platform_profile.c
iwanders: Hmm, the current [`convert_profile_to_ssam`](https://github.com/linux-surface/kernel/blob/8cd23b1bb3a8b7a3ef7cec2c37e7e46e6397a858/drivers/platform/surface/surface_platform_profile.c#L104) function uses `EOPNOTSUPP`, so I foolowed that, I'm fine with changing it over, but it feels like they should both return the same error.

drivers/platform/surface/surface_aggregator_registry.c
iwanders: yeah, that's a good idea actually will change that over.

qzed: Just to mention it here as well: I somehow misremembered and though the fan was in the TMP category, not FAN. So disregard the part mentioning the fan monitor patches. That's all fine.

iwanders: So what naming convention do you want to follow if it has two subsystems? :D 

If we use the subsystem naming convention, I guess we end up with: 
- `ssam_node_tmp_perf_profile` (instead the original `ssam_node_tmp_pprof`)
- `ssam_node_tmp_fan_perf_profile` (the new thing, currently named `ssam_node_perf_profile_with_fan`)

qzed: Let's stick with `tmp` for this one (i.e. `ssam_node_tmp_perf_profile`), since it's still kind of the main thing.

iwanders: So for the non-fan version `ssam_node_tmp_perf_profile` and for the fan version `ssam_node_tmp_perf_profile_with_fan`?

I'm actually not sure what the TMP profile actually does... we should check some clock speeds or something to determine if it does anything at all >_<

qzed: Yeah, looks good.

On the SB2 it changes the fan profile of the dGPU fan. But that's about all that I could figure out. I doubt it changes any clocks. SAM shouldn't have access that direct to the CPU, I think.

drivers/platform/surface/surface_platform_profile.c
qzed: Oh hmm, you're right. Then we should stick with `EOPNOTSUPP`.

drivers/platform/surface/surface_aggregator_registry.c
iwanders: > SAM shouldn't have access that direct to the CPU, I think.

Hmm, I'm not so sure of that, I found the string `IMVP MP2964 (U6601)` in the firmware blob for it... which definitely looks like [this chip](https://www.monolithicpower.com/en/documentview/productdocument/index/version/2/document_type/Datasheet/lang/en/sku/MP2965/document_id/5134/) that is specificially made to control power to the CPU, and I wouldn't be surprised if one of the 'low power' modes drops the voltage of the cpu down a bit to force it to reduce the clock speed? Like that would work on a microcontroller at least, I'm not sure if that  can be done on a modern CPU...

There's also `Batt Charger BQ25713 (U6302)` in the firmware, which seems like an interesting [IC](https://www.ti.com/lit/ds/symlink/bq25713.pdf) as it provides access to various currents, which would allow some insight into whether low power mode is actually lower power... I tried to find that, but just loading `i2c-dev` resulting in a bunch of i2c busses, but none of them (except the display one if I remember correctly) had anything on it that answered back. I part expect that we don't actually have access to the useful i2c bus... because that's what the management ic is for :(

iwanders: Incorporated.

drivers/platform/surface/surface_platform_profile.c
```diff
@@ -14,67 +14,93 @@
 
 #include <linux/surface_aggregator/device.h>
 
-enum ssam_tmp_profile {
-	SSAM_TMP_PROFILE_NORMAL             = 1,
-	SSAM_TMP_PROFILE_BATTERY_SAVER      = 2,
-	SSAM_TMP_PROFILE_BETTER_PERFORMANCE = 3,
-	SSAM_TMP_PROFILE_BEST_PERFORMANCE   = 4,
+// Enum for the platform performance profile sent to the TMP module.
+enum ssam_perf_profile {
```
iwanders, line 18: I'm actually wondering now whether we should keep `tmp` in this name as well... so just keep this as it was; `ssam_tmp_profile` and introduce the `ssam_fan_profile`. The device struct would still be the more generic `ssam_platform_profile_device`?

Naming is hard if we don't really know what these things mean or do :(

drivers/platform/surface/surface_aggregator_registry.c
```diff
@@ -68,12 +68,26 @@ static const struct software_node ssam_node_bat_sb3base = {
...
+ */
+static const struct property_entry ssam_node_perf_profile_has_fan[] = {
+	PROPERTY_ENTRY_BOOL("has_fan"),
+	{ }
+};
+
+static const struct software_node ssam_node_tmp_perf_profile_with_fan = {
+	.name = "ssam:01:03:01:00:01",
+	.parent = &ssam_node_root,
+	.properties = ssam_node_perf_profile_has_fan,
```
qzed, line 88: Not sure if that's a good idea, but apparently it's possible to directly put the properties here instead of defining an external array. [`[HWMON_CHANNEL_INFO()`](https://elixir.bootlin.com/linux/v6.6.9/C/ident/HWMON_CHANNEL_INFO)`](https://elixir.bootlin.com/linux/v6.6.9/source/include/linux/hwmon.h#L428) does something like that. Would reduce the amount of long names by one.

drivers/platform/surface/surface_platform_profile.c
qzed: Yeah. I'd vote for `ssam_perf_profile` and `ssam_fan_profile`. I think that best reflects what we know (so far). No need to add the `tmp` everywhere. I think it might be good to keep in direct SAM interface function definitions, but we don't need it for the rest. Especially if we separate into those two categories.

qzed: On the other hand we could also do `ssam_tmp_profile` and `ssam_fan_profile` and separate them by the target category.

drivers/platform/surface/surface_aggregator_registry.c
qzed: > > SAM shouldn't have access that direct to the CPU, I think.
> 
> Hmm, I'm not so sure of that, I found the string `IMVP MP2964 (U6601)` in the firmware blob for it... which definitely looks like [this chip](https://www.monolithicpower.com/en/documentview/productdocument/index/version/2/document_type/Datasheet/lang/en/sku/MP2965/document_id/5134/) that is specificially made to control power to the CPU, and I wouldn't be surprised if one of the 'low power' modes drops the voltage of the cpu down a bit to force it to reduce the clock speed? Like that would work on a microcontroller at least, I'm not sure if that can be done on a modern CPU...

Hmm interesting... The reason I doubted that SAM would have access is that there's already a software-interface for controlling the clock speeds/power limits on modern CPUs. I would have assumed that Windows uses that rather than go through SAM somehow.

iwanders: Yeah, that's probably the 'main' thing, but then the question is what does the whole TMP module still do when we switch profiles :thinking:

drivers/platform/surface/surface_platform_profile.c
iwanders: Yeah, I think
> On the other hand we could also do ssam_tmp_profile and ssam_fan_profile and name them by the target category.

Is the best to go with for now tbh, especially since we don't have any further information on the function of the tmp profile setting.

drivers/platform/surface/surface_aggregator_registry.c
iwanders: Hmm, I tried this:

```c
	.properties = (((struct property_entry[]) {
		PROPERTY_ENTRY_BOOL("has_fan"),
		{}
	})),
```
But that doesn't compile because the `PROPERTY_ENTRY_BOOL` return is not considered const?
```
surface_aggregator_registry.c:89:51: error: initializer element is not constant
   89 |         .properties = (((struct property_entry[]) {
      |                                                   ^
```

I'd I'm not sure if this would be better either. The current is clear and readable, though a bit verbose. I renamed it from `ssam_node_perf_profile_has_fan` to  `ssam_node_tmp_perf_profile_has_fan` to have the `tmp` prefix in there for consistency.

drivers/platform/surface/surface_platform_profile.c
iwanders: `ssam_tmp_profile` is back in place everywhere, delta is much smaller now.

drivers/platform/surface/surface_aggregator_registry.c
qzed: Yeah, I was mostly thinking that we maybe could reduce a bit of verbosity and make it look a bit nicer, given that that property is only used once. It should work if you use
```c
	.properties = (const struct property_entry[]) {
		PROPERTY_ENTRY_BOOL("has_fan"),
		{}
	},
```
The `properties` field is const (and you don't need those two outer braces).

But I'm also not sure if upstream is okay with that, at least I haven't found any instance of this being done before.

iwanders: Copy pasting the proposed code also gives:
```
surface_aggregator_registry.c:83:55: error: initializer element is not constant
   83 |         .properties = (const struct property_entry[]) {
      |                                                       ^
```

I'm fine with keeping this as is, given that I'm not sure how to express this in a way the compiler likes :P

I've changed the name to be `ssam_node_tmp_perf_profile_has_fan` for consistency with having the `tmp` in there.

StollD: The way I understand it, values cannot be const, only variables. So the const in this case is a NOP.

qzed: Hmm, this works for me when building as an external module and when building from `v6.6-surface`. Now that I'm checking it properly, it also works without the `const`... no clue why it doesn't work for you. Anyway, we can keep it as external array.

iwanders: Well, it may be a compiler issue? 

I made a minimal (non-)working example. With gcc 12.3 (and down to  in [godbolt](https://godbolt.org/z/rncnWcjvK), it reproduces my error:
```
<source>:61:51: error: initializer element is not constant
   61 |     .properties = (const struct property_entry[]) {
      |                                                   ^
<source>:61:51: note: (near initialization for 'our_node_nonconst_const')
Compiler returned: 1
```

But with gcc 13.1 it compiles fine [godbolt](https://godbolt.org/z/afYcbxsTf).

I started pruning compilation flags, but even until [gcc 10.1](https://godbolt.org/z/nvMdh8Kah) it won't compile...

Not sure what is happening, but it could explain why we don't see such a construct anywhere else in the codebase? Could be that one of the flags causes this error? This is just the default flags I get when compiling on NixOS.

qzed: Huh, yeah. No idea why it works on one but not the other. Maybe some C standard stuff? Thanks for checking that. Let's keep it like it is then. I'll try to do a proper review some time in the next days.

iwanders: Friendly ping, trying to finish up my tablet setup and also wrap up outstanding PRs around it :)

If you're happy with this I'll rebase it on Torvald's `master` branch and send it to the `platform-driver-x86` list.

qzed (approved):

drivers/platform/surface/surface_aggregator_registry.c
qzed: Sorry! Had planned to take it up this weekend but a conference deadline got extended from march 1 until march 15 and things are currently a bit stressful in my day-job. Just had another quick look and I think it's ready.

iwanders: All good! I know open source stuff is best effort :) Hope your conference submission is going well.

I'll aim to get this sent upstream this weekend. Thanks for the review.

qzed:
@StollD if you have some time, feel free to merge this and add it to the patches. Otherwise I'll try to make some time, but I'm not sure I can get to much before March 15th.

iwanders:
Upstream lore [entry](https://lore.kernel.org/platform-driver-x86/20240302170147.13026-1-ivor@iwanders.net/T/) for v1.

StollD:
I cherry-picked this to our 6.6 and 6.7 branches. Thank you!

PR #146, opened 2024-02-03
Fix camera issues in 6.7.x on SGo2

mojyack:
After linux-6.7(or 6.6?), ipu3-imgu and dw9719 are broken. See https://github.com/linux-surface/linux-surface/issues/1323.
This patches fix those issues.
Tested on Surface Go 2.

PR #149, opened 2024-03-25
Use upstream dw9719 driver

mojyack:
In #146, I reverted upstream dw9719 driver to the old one.
This PR allows using upstream driver on surface go 2.


qzed:
Thanks!

PR #151, opened 2024-04-19
hid: ithc: Update from quo/ithc-linux

StollD:
This enables the touchscreen and the touchpad on the Surface Laptop Studio 2. See also here: https://github.com/linux-surface/linux-surface/issues/1272

The touchpad will behave a little bit weird, because iptsd treats it as a touchscreen. But I think I have everything figured out to make it create an actual touchpad device, just need to write the code. And having the driver work out of the box will help with testing.


qzed (approved):
The code looks like it could use a bit of clean up, but I think that's something that we can do later. Seeing as this gets us one step closer to a working touchpad, I think it should be merged. So feel free to merge it.

PR #152, opened 2024-04-21
platform/surface: aggregator: Defer probing when serdev is not ready

phreer:
Failure of probing of SSAM serial hub was observed (https://github.com/linux-surface/linux-surface/issues/1271) from time to time due to race condition where the underlying uart device could be not ready at the point calling serdev_device_open(). Specifically, the failing path is

```
ssam_serial_hub_probe()
    => serdev_device_open()
        => ctrl->ops->open() // this is ttyport_open()
            => tty->ops->open() // this is uart_open()
                => tty_port_open()
                    => port->ops->activate() // this is uart_port_activate()
                        Find bit UPF_DEAD is set in uport->flags and fail with errno -ENXIO.
```

Retrying upon failure in opening serdev device would greatly alleviate this situation.  Tested with my Surface Pro 5 rebooting for 10+ times and didn't see the issue anymore.

I admit this solution might be suboptimal, but it does help cure the issue that users lost their battery indicator or touchpad functionality randomly.

### DIFF
--- a/drivers/platform/surface/aggregator/core.c
+++ b/drivers/platform/surface/aggregator/core.c
@@ -645,9 +645,20 @@ static int ssam_serial_hub_probe(struct serdev_device *serdev)
 	/* Set up serdev device. */
 	serdev_device_set_drvdata(serdev, ctrl);
 	serdev_device_set_client_ops(serdev, &ssam_serdev_ops);
+
+	/* The following step can fail when it's called too early before the
+	 * underlying uart device is ready (in this case -ENXIO is returned).
+	 * Instead of simply giving up and losing everything, we can defer
+	 * the probing by returning -EPROBE_DEFER so that the kernel would be
+	 * able to retry later. */
 	status = serdev_device_open(serdev);
-	if (status)
+	if (status == -ENXIO)
+		status = -EPROBE_DEFER;
+	if (status) {
+		dev_err_probe(&serdev->dev, status,
+			      "failed to open serdev device\n");
 		goto err_devopen;
+	}
 
 	astatus = ssam_serdev_setup_via_acpi(ssh->handle, serdev);
 	if (ACPI_FAILURE(astatus)) {
--- a/drivers/platform/surface/aggregator/core.c
+++ b/drivers/platform/surface/aggregator/core.c
@@ -623,8 +623,10 @@ static int ssam_serial_hub_probe(struct serdev_device *serdev)
 	acpi_status astatus;
 	int status;
 
-	if (gpiod_count(&serdev->dev, NULL) < 0)
+	if (gpiod_count(&serdev->dev, NULL) < 0) {
+		dev_err(&serdev->dev, "no GPIO found\n");
 		return -ENODEV;
+	}
 
 	status = devm_acpi_dev_add_driver_gpios(&serdev->dev, ssam_acpi_gpios);
 	if (status)
@@ -637,8 +639,11 @@ static int ssam_serial_hub_probe(struct serdev_device *serdev)
 
 	/* Initialize controller. */
 	status = ssam_controller_init(ctrl, serdev);
-	if (status)
+	if (status) {
+		dev_err_probe(&serdev->dev, status,
+			      "failed to initialize ssam controller\n");
 		goto err_ctrl_init;
+	}
 
 	ssam_controller_lock(ctrl);
 
@@ -663,6 +668,7 @@ static int ssam_serial_hub_probe(struct serdev_device *serdev)
 	astatus = ssam_serdev_setup_via_acpi(ssh->handle, serdev);
 	if (ACPI_FAILURE(astatus)) {
 		status = -ENXIO;
+		dev_err(&serdev->dev, "failed to setup serdev\n");
 		goto err_devinit;
 	}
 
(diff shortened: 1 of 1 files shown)


qzed:
Thanks for the PR (and especially the breakdown of what goes wrong). But as you said, I think we can do better: Generally, in cases like this the driver should return `EPROBE_DEFER`. This tells the device driver core that the driver is not yet ready and missing some dependency, and that it should try again later. Could you try something like this?
```c
if (status == -ENXIO)  // or whatever is returned by serdev_device_open() when it fails due to not being ready
	return -EPROBE_DEFER;
else if (status)
	return status;
```

Apart from me not knowing the specific error code, the reason I haven't added this yet is that I don't know where specifically this should be addressed. I.e., should `serdev_device_open()` return `EPROBE_DEFER`  or should we check the error code and do it ourselves in the SAM driver?

Due to that, my current suggestion for a workaround would be to build-in the required modules (`8250_dw` and pinctrl ones) and avoid the race condition in that way (like Fedora and Ubuntu/Debian already do). This seems to have worked [here](https://github.com/linux-surface/linux-surface/issues/1319#issuecomment-2059468939) and I was planning on doing that today. Right now, I'm also not sure whether this could also fail due to pinctrl dependencies not being present. Some users have reported that the buttons (soc-button-array) on Arch do not work due to that. And if pinctrl isn't there, the `gpiod_count()` or `ssam_irq_setup()` calls might fail too. So we probably want to build in at least those anyways.

If I remember correctly Hans de Goede mentioned that (on x86?) the pinctrl stuff should be always built in. And since I'm already mentioning him:

@jwrdegoede: Maybe you know what the proper way of returning `EPROBE_DEFER` above is?

qzed:
@phreer: I've opened a PR for building in the modules at https://github.com/linux-surface/linux-surface/pull/1426. In case we merge this before you get around to testing with `EPROBE_DEFER`, please make sure that you do not build in those modules.

phreer:
So `serdev_device_open()` returning error is only part of the story, there might be somewhere else we get an error. OK, looks like https://github.com/linux-surface/linux-surface/pull/1426 is a more reasonable workaround for the time being. I'd love to investigate how to return `EPROBE_DEFER` thing properly and figure out a long-term solution for this issue :).

qzed:
I think the proper way is probably a mix of both: Build in the pinctrl modules or have them in the initramfs (as a lot of stuff relies on them) but don't do that for the serial ones as they are a bit more specific. Then return `EPROBE_DEFER` (probably in the SAM driver) if that works to make it re-try if the serial driver isn't there yet.

phreer:
I just got basic idea how deferred driver probing works in kernel and will play around with it soon, hopefully enable this functionality for SAM driver.

But I don't get the point why

qzed:
> But I don't get the point why we still require modules like pinctrl to be built in even if deferred probing is supposed.

Returning `EPROBE_DEFER` for `serdev_device_open()` only helps for the serial driver. So anything that relies on the pinctrl drivers being present might still fail.

phreer:
Yeah, of course. But what if we also return `EPROBE_DEFER` when resources like GPIO are not ready during `ssam_serial_hub_probe()`? IIUC, your point is that we can hardly get an exhaustive list of cases where we need defer probing instead of returning real error. Maybe we could just include all these cases gradually, starting from the most obvious ones.

qzed:
@phreer 

> Yeah, of course. But what if we also return `EPROBE_DEFER` when resources like GPIO are not ready during `ssam_serial_hub_probe()`? IIUC, your point is that we can hardly get an exhaustive list of cases where we need defer probing instead of returning real error. Maybe we could just include all these cases gradually, starting from the most obvious ones.

I think that would be the most ideal solution. But pinctrl drivers are spread out quite a bit, so I'm not really sure what all would need to be done/checked here.

qzed:
@jwrdegoede 

> As for this specific case, where does the serdev the driver tries to open come from ? I would expect this to be created by the tty subsystem, specifically by `acpi_serdev_register_devices()` from `drivers/tty/serdev/core.c` based on some ACPI device having a UartSerialBus resource pointing to the port.
> 
> Since this serdev is instantiated at the time the tty subsystem is probing it I find it weird that when your serdev driver (I assume you are using a serdev driver?) tries to bind to it that it is not ready yet.

It is a serdev driver, yes. From the reports I've got, building in `8250_dw` and the intel LPSS drivers does provide a functioning workaround. So due to that, to me it looks like there is some stuff going on in parallel. Like some parts of serdev are registered so that our driver can probe, but the underlying provider of the port is not yet ready. I know too little of the underlying things at work here to say what exactly is happening though.

I will try to look into it a bit more and then send a mail to linux-serial. Thanks!

phreer:
Hi @qzed, me again. As you suggested, I tried the way of deferring probing and it worked as expected.

Only deal with serdev case for now. Let's be patient and identify more race condition points in the future.

qzed (commented):
I think it would look a bit nicer if you do something like
```c
if (status == -ENXIO)
	status = -EPROBE_DEFER;
if (status) {
	ssam_err_probe(...);
	goto err_devopen;
}
```
You'd have to add `ssam_err_probe()` similar to the other print macros though with `dev_err_probe()` as underlying print function.

phreer:
@qzed Sure! Do you think it is a good idea to add more messages upon error, e.g., gpiod_count?

qzed (approved):
Thanks! Looks good to me.

phreer:
Hope these error messages would help!

phreer:
Hi @qzed ,  I've completed my code and would appreciate guidance on how to proceed further.

qzed:
Hi, sorry for the delay. I've been a bit busy with day-time work the last days again.

I'm not a big fan of moving the controller init to the front, I'd prefer if we can keep the order (with smaller/lighter init stuff in the beginning). Just use the `dev_` functions for logging with that. And thinking about that: I guess we also don't really need the `ssam_err_probe()` because `serdev->dev` is directly available here. Sorry for not noticing that earlier.

Apart from that I think it looks good.

I think you could also submit the `EPROBE_DEFER` patch as RFC to the pdx86 and linux-serial mailing lists to get the discussion started on that, if you want. I'm not sure when I will find the time for that, as I will be away for a large part of May and have some stuff to take care off before.

phreer:
Sure, it makes sense to let light staffs go first so we can fail fast. I'm pleased to start a discuss the EPROBE_DEFER patch on mailing list. Lastly, I appreciate your assistance despite your busy schedule. Thank you!

qzed (approved):
Thanks for sticking with it! Looks good to me now.

PR #154, opened 2024-05-14
Make surface_fan_hwmon_is_visible static to avoid compiler warning.

jasisonee:
Compiling with CONFIG_WERROR enabled fails because of -Wmissing-prototypes.


qzed (approved):
Thanks!

PR #158, opened 2024-08-25
Added patch to allow optional hibernate during lockdown

jeduardo:
Expose an optional lockdown_hibernate kernel parameter to enable hibernation during kernel lockdown mode. The patch was imported from [here](https://gist.github.com/brknkfr/95d1925ccdbb7a2d18947c168dfabbee).

Closes https://github.com/linux-surface/linux-surface/issues/1543

### DIFF
--- a/Documentation/admin-guide/kernel-parameters.txt
+++ b/Documentation/admin-guide/kernel-parameters.txt
@@ -3000,6 +3000,11 @@
 			to extract confidential information from the kernel
 			are also disabled.
 
+	lockdown_hibernate	[HIBERNATION]
+			Enable hibernation even if lockdown is enabled. Enable this only if
+			your swap is encrypted and secured properly, as an attacker can
+			modify the kernel offline during hibernation.
+
 	locktorture.acq_writer_lim= [KNL]
 			Set the time limit in jiffies for a lock
 			acquisition.  Acquisitions exceeding this limit
--- a/kernel/power/hibernate.c
+++ b/kernel/power/hibernate.c
@@ -37,6 +37,7 @@
 #include "power.h"
 
 
+static int lockdown_hibernate;
 static int nocompress;
 static int noresume;
 static int nohibernate;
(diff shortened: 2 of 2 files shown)


jeduardo:
@qzed just a heads up that the the patch is here. I raised it against the latest surface-devel branch as per the instructions. Please let me know if I need to change anything else.

qzed:
Yeah, I saw, thanks! I just didn't get around to look at it yesterday. I'll try to have a look at it and merge it this evening.

qzed (approved):
Thanks!

qzed:
Sorry that it took me a while, but I finally managed to update the patches and prepare a new release.

jeduardo:
Hey @qzed, thanks for the update! I've seen the packages were built in the pipeline but it seems they did not yet land into the main repo (tried an update but got no updated versions for `kernel-surface`. If you need any help with testing let me know!

qzed:
Should be in the repo now.

jeduardo:
@qzed, it indeed is and I just downloaded it. I can also confirm that the patch is working as expected and I can now hibernate while SecureBoot is on. Thanks!

mirko:
I just applied that patch to a vanilla linux kernel (6.17) to allow hibernation while still being in lockdown:

maybe I'm missing something, but it seems my kernel does /not/ anymore automatically start into lockdown at all with this patch applied.
There's no lockdown-info in dmesg, `/sys/kernel/security/lockdown` states `[none]` and I can happily open `/dev/mem` from userspace.

When starting with `cmdline` param `lockdown=integrity`, though, lockdown is definitely enabled. But now I'm also unable to hibernate anymore, due to lockdown.

Is this patch based on other patches? Does it require certain config options?

```
$ grep -E '(LOCKDOWN|SECURE.*BOOT)' .config
CONFIG_SECURITY_LOCKDOWN_LSM=y
CONFIG_SECURITY_LOCKDOWN_LSM_EARLY=y
CONFIG_IMA_SECURE_AND_OR_TRUSTED_BOOT=y
```

Can someone double-check, if their kernel is actually locked down with this patch applied?

PR #159, opened 2025-02-23
Add support for Surface Pro 11

lukas-hetzenecker:


qzed:
Thanks! Would you be willing to also send this in upstream? I can give you some guidance on how to do this (the gist is described [here](https://docs.kernel.org/process/1.Intro.html); alternatively I can also send it in if you prefer).

lukas-hetzenecker:
Hello @qzed - sure, I'm happy to send it upstream

christian-bendiksen:
Will a patch like this work for the Surface Laptop 7 (snapdragon x elite)?

This is the ID i could find for the Touchpad: MSHW0238

qzed:
@christian-bendiksen ARM/QCOM devices need device trees at the moment, so unfortunately it's not that easy. There is already some limited support in the kernel, but you'll probably want to compile a custom kernel for now. See https://github.com/linux-surface/linux-surface/issues/1590.

PR #164, opened 2026-05-15
Patch enabling speakers, mic, camera, suspend and hibernate, and touchpad configuration for surface pro 11 intel.

AndreGilerson:
As requested in https://github.com/linux-surface/linux-surface/pull/1990 a separate PR for the kernel changes. These changes enable audio, camera, suspend/hibernate support on the surface pro for business 11th edition with intel. 

Please refer to https://github.com/linux-surface/linux-surface/pull/1990 for the contrib packages that perform additional setup outside the kernel for audio and camera specifically. 

Tested with a clean ubuntu 26.04 install. But additional testing is welcome by  other users.


zR-JB:
@AndreGilerson I opened linux-surface/kernel#176 as a draft/RFC based substantially on your IMX681 work here, with the goal of converging rather than maintaining a separate SP12 driver.

I would prefer if we could eventually split the IMX681/front-camera work into its own focused PR and compare the driver in #176 with your SP11 version. My SP12 version strips out a number of things that I could not verify as necessary on SP12 and keeps only the tested clock/power/reset setup, fixed mode/timing and raw analogue-gain control.

Would you be able to check whether that reduced version also works on SP11? If it does, I'd be interested whether you see any reason to retain the pieces I removed, or whether we could use the smaller driver as the common SP11/SP12 implementation.

I have kept #176 as a draft specifically so we can work out the common implementation and attribution/DCO together first.

